PR #1, opened 2025-07-02
[Lotto - Step 1] Submit Deniz' mission

deniz-oezdemir:
This pull request adds a console-based lotto game. It covers the core game logic, input and output handling, and error management. The changes affect several files and focus on keeping the code modular and maintainable.

We would especially appreciate feedback on the following questions:
- Is our file structure well-suited to the program’s complexity?
- Are the breadth and depth of the tests adequate?
- Should we add more documentation, such as comments, more explanations in the README, or other materials?
- Are we following best practices regarding development with kotlin?

This implementation is ready for review. I would greatly appreciate your feedback.

PR #2, opened 2025-07-03
[Lotto - Step 1] Submit codebrew42's mission

jieun-pak:
This pull request introduces a console-based Lotto program.
It includes the core game logic, input/output handling, and error management.
Our team tried to adapt to Kotlin conventions, but we know there’s room for improvement.

I’d appreciate feedback on:
1. Overall code structure: Would it make sense to split parts into additional files or modules for better clarity or maintainability?
2. Testing strategy: Any critical improvements or fixes for the tests?
3. Kotlin style: Tips on better following Kotlin's convention.

I’m also open to suggestions for improving the testing strategy to ensure the program is reliable.
Thank you in advance for your insights and help!

— Jin Park


jieun-pak:
Step 1 #1

README.md
```diff
@@ -1 +1,32 @@
...
+
+## features
+
+The features were implemented in the following order:
+
+- [x] get input from user: purchase amount from user
+  - [x] units of 1000
+  - [x] greater than or equal to 1000
+- [x] display purchase amount `N`
+- [x] display N-Sets of randomly generated numbers
```
pkeugine, line 15: It's nice to have detailed feature list. 👍  
I think detailed documents help teammates and myself  
understand what's going on with the project as we maintain it.  

Also, the commit history seems like it's based on this list.
Nice work 💯

src/main/kotlin/Application.kt
```diff
@@ -0,0 +1,8 @@
+import controller.Controller
+import view.InputView
+import view.OutputView
+
+fun main(){
+    val controller = Controller()
+    controller.run(inputView = InputView(), outputView = OutputView())
+}
```
pkeugine: image

When reviewing the code, I noticed this red indicator at the end of some files.
Some files have it, while others don't. I'd like you to find the difference between them.

Additionally, it would be helpful to find out why having this newline is preferred.  
**Hint**: [it's related to missing empty lines](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file) 😉  

In IntelliJ, there's an option to automatically add a newline at the end of a file. It’s worth checking that out.

gradlew.bat
```diff
@@ -1,92 +1,92 @@
...
+@rem ##########################################################################
+
+@rem Set local scope for the variables with windows NT shell
+if "%OS%"=="Windows_NT" setlocal
+
+set DIRNAME=%~dp0
+if "%DIRNAME%"=="" set DIRNAME=.
+@rem This is normally unused
+set APP_BASE_NAME=%~n0
+set APP_HOME=%DIRNAME%
```
pkeugine: Woa, this file looks kinda hard.  
I'm sure you didn't write this file, but it's in the pull request.  
What does this `gradlew.bat` file do?

build.gradle.kts
```diff
@@ -18,6 +18,7 @@ dependencies {
     testImplementation("org.junit.jupiter", "junit-jupiter", "5.10.2")
     testImplementation("org.assertj", "assertj-core", "3.25.3")
     testImplementation("io.kotest", "kotest-runner-junit5", "5.8.0")
+    testImplementation(kotlin("test"))
```
pkeugine, line 21: I think it’s important for us as developers to maintain the tools we use and only import what we truly need, even for writing tests. It’s good practice to understand what we’re using.

That makes me curious: 
what were the reasons that led you and your pair to add `testImplementation(kotlin("test"))`?
I’d like you to look into what it does, how it compares to other libraries (like AssertJ), and explain it to me. 😊

src/main/kotlin/view/InputView.kt
```diff
@@ -0,0 +1,104 @@
...
+        private const val GET_PURCHASE_AMOUNT = "Please enter the purchase amount.\n"
+        private const val GET_WINNING_NUMBERS = "Please enter last week's winning numbers.\n"
+        private const val GET_BONUS_NUMBERS = "Please enter last week's bonus number.\n"
+        private const val INVALID_INPUT_MESSAGE = "The input is invalid!"
+        private const val RETRY_INPUT_MESSAGE = "Enter the input again!"
+        private const val ERROR_INVALID_DIGITS = "[ERROR] Winning numbers must be valid digits."
+        private const val ERROR_BONUS_RANGE =
+            "[ERROR] Number must be between ${Lotto.TICKET_NUMBER_MINIMUM} and ${Lotto.TICKET_NUMBER_MAXIMUM}."
+        private const val ERROR_DUPLICATE = "[ERROR] Numbers can't be duplicates."
+        private const val ERROR_NUMBERSET_SIZE = "[ERROR] There must be ${Lotto.TICKET_LENGTH} numbers."
```
pkeugine: I see that you've made both  
* application console message
* and error message  

constants, keeping them together.  

It looks nice, but whenever I want to understand what a message says while reading through the logic,
I have to scroll down here, search for it, and then go back up to continue.
I wouldn't say this is the wrong way to organize messages,
but personally, I find it a bit less convenient compared to having them closer to the logic.

I'm curious about your experience, how did it feel for you?

src/main/kotlin/view/InputView.kt
```diff
@@ -0,0 +1,104 @@
...
+        while (true) {
+            print(prompt)
+            val input = readLine()
+            if (input != null) {
+                return input
+            } else {
+                println(RETRY_INPUT_MESSAGE)
+            }
+        }
+    }
```
pkeugine: `while (true)` can be a dangerous move. It often leads to bugs no one intended.  
It may look unharmful for now, but trust me, I've spent a lot of Friday nights with this guy 🥲 

Well, yes, I totally get that using a `while (true)` loop to keep asking for valid user input is a classic and valid use case, and many developers advocate for it.

Still, there are other options available to us, so it’s worth thinking about what those are, and whether they might be a better fit than a `while (true)` loop 🤓

src/main/kotlin/view/InputView.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    private fun parsePurchaseAmount(input: String): Int? {
+        return input.toIntOrNull()
+    }
```
pkeugine: image

Some users won't behave normally(?) like we hope them to.  
But I think we can cover above situation by trimming the whitespace before parsing them to Int.  
Let's try it :)

src/main/kotlin/view/InputView.kt
```diff
@@ -0,0 +1,104 @@
...
+            } catch (e: IllegalArgumentException) {
+                println(e.message)
+            }
+        }
+    }
+
+    fun validateBonusNumber(bonusNumber: Int, winningNumbers: List<Int>) {
+        require(bonusNumber in Lotto.TICKET_NUMBER_MINIMUM..Lotto.TICKET_NUMBER_MAXIMUM) { ERROR_BONUS_RANGE }
+        require(!winningNumbers.contains(bonusNumber)) { ERROR_DUPLICATE }
+    }
```
pkeugine: Getting the input for the winning lotto numbers and bonus number in `InputView` looks great.
However, handling validation here feels like a bit too much responsibility for `InputView`.
Why don’t we separate that concern into dedicated classes for lotto numbers and let them handle the validation?

src/test/kotlin/model/LottoTest.kt
```diff
@@ -0,0 +1,65 @@
...
+
+import org.junit.jupiter.api.Assertions.assertDoesNotThrow
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import kotlin.test.assertTrue
+
+class LottoTest {
+    @Test
+    fun `invalid purchase amount not divisible by unit`() {
```
pkeugine:
```suggestion
    fun `exception : invalid purchase amount - not divisible by unit`() {
```

This is just my opinion, but formatting test names consistently really helps with maintaining and writing new tests as a team.
The suggestion above is just an example. It’d be great for you and your pair to figure out what works best for you.

src/main/kotlin/model/Statistics.kt
```diff
@@ -0,0 +1,29 @@
...
+                Rank.FIRST -> matches[0] += 1
+                Rank.SECOND -> matches[1] += 1
+                Rank.THIRD -> matches[2] += 1
+                Rank.FOURTH -> matches[3] += 1
+                Rank.FIFTH -> matches[4] += 1
+                Rank.MISS -> matches[5] += 1
+            }
+        }
+        return matches
+    }
```
pkeugine: There are many collection types and tools we can use for storing data.
You used a `MutableList`, but I noticed that `IntArray` could also be an option here.
What are the differences between the two?

Also, since the list is mutable, returning `matches.toList()` might be a better option.
Why do you think that is?

**TIP**: Using `Cmd + B` (or `Ctrl + B` on Windows) with your cursor on a class or method will take you to its source code.
It’s super helpful for understanding how things work. I use it all the time. :)

src/main/kotlin/model/Lotto.kt
```diff
@@ -0,0 +1,37 @@
...
+        }
+        ticket.sort()
+        return ticket
+    }
+
+    companion object {
+        const val PURCHASE_AMOUNT_UNIT = 1000
+        const val TICKET_LENGTH = 6
+        const val TICKET_NUMBER_MINIMUM = 1
+        const val TICKET_NUMBER_MAXIMUM = 45
```
pkeugine: Just like I mentioned in the InputView comment, 
separating concerns by creating dedicated classes can make the project easier to maintain.
For example, the `tickets` field could hold `LottoTicket`s that each contains a list of `LottoNumber`s:

```kotlin
data class LottoNumber(val value: Int)

class LottoTicket(val numbers: List<LottoNumber>)
```

By doing so, we'll be able to move the validation and these magic numbers to them.

pkeugine: I also recommend trying to make all the `values` (ex: lotto numbers) into classes and keeping the validation in it. It helps us think about separating responsibilities as our project gets bigger.

src/main/kotlin/model/Rank.kt
```diff
@@ -0,0 +1,25 @@
+package model
+
+enum class Rank(val countOfMatch: Int, val winningMoney: Int) {
+    FIRST(6, 2_000_000_000),
+    SECOND(5, 30_000_000),
+    THIRD(5, 1_500_000),
+    FOURTH(4, 50_000),
+    FIFTH(3, 5_000),
+    MISS(0, 0);
```
pkeugine: Nice job using an enum class for ranking system. 👍 
You've used classes, enum, an object (`Statistics`), and even a class that only contains a function (`Application`).  
What are the differences, and could there be more?

**Hint**: [data class](https://kotlinlang.org/docs/data-classes.html#)

src/main/kotlin/view/InputView.kt
```diff
@@ -0,0 +1,104 @@
...
+            } else {
+                println(INVALID_INPUT_MESSAGE)
+            }
+        }
+    }
+
+    private fun convertWinningNumbers(input: String): List<Int> {
+        return try {
+            input.split(",")
+                .map { it.trim().toInt() }
```
pkeugine:
```suggestion
            input.split(",")
                .map(String::trim)
                .map(String::toInt)
```

Kotlin offers A LOT of ways for us to do things, and readability doesn't always have one correct answer.
So this review is just a suggestion. Which one looks better?

src/main/kotlin/model/Lotto.kt
```diff
@@ -0,0 +1,37 @@
+package model
+
+class Lotto(val purchaseAmount: Int) {
+    val numberOfTickets: Int
+    val tickets = mutableListOf<MutableList<Int>>()
```
pkeugine: Yet another review suggesting class separation 😂

I’d recommend creating a dedicated domain class for `tickets`, and letting that class be the **only one** responsible for mutating the list.
It’s too risky to let other parts of the code access a mutable list, especially when it represents something important like lotto tickets. 💸 💸 

An easy way to think about it:
**Is it mutable?** Then let as few parts of the code touch it as possible.

By separating classes, it'll be easier to test them too!

src/test/kotlin/model/LottoTest.kt
```diff
@@ -0,0 +1,65 @@
...
+            Lotto(10)
+        }
+    }
+
+    @Test
+    fun `invalid purchase amount negative`() {
+        assertThrows<IllegalArgumentException> {
+            Lotto(-10)
+        }
+    }
```
pkeugine: How was your TDD experience? I’d love to hear how it helped you, and which parts felt challenging.

src/test/kotlin/model/LottoTest.kt
```diff
@@ -0,0 +1,65 @@
...
+    fun `ticket numbers are in valid range 1 to 45`() {
+        val lotto = Lotto(1000)
+        val ticket = lotto.fillTicket()
+
+        for (number in ticket) {
+            assertTrue(number >= 1 && number <= 45, "Number $number is not in valid range 1-45")
+        }
+    }
+
+
```
pkeugine: I see unnecessary empty line 👀

src/test/kotlin/model/LottoTest.kt
```diff
@@ -0,0 +1,65 @@
...
+    @Test
+    fun `ticket numbers are in valid range 1 to 45`() {
+        val lotto = Lotto(1000)
+        val ticket = lotto.fillTicket()
+
+        for (number in ticket) {
+            assertTrue(number >= 1 && number <= 45, "Number $number is not in valid range 1-45")
+        }
+    }
```
pkeugine: Excellent job testing the lotto ticket. 👍👍
However, as a developer, I’d like to manage the ticket numbers myself so I can create a ticket with exactly the numbers I want.
This would give me more control over the tests. How could we achieve that? 🤔

pkeugine (changes requested):
Hello Jin 🙌
The project looks great, and I can see you’ve managed your README file very well. 👍

I left some comments where things can be improved, along with a few suggestions that are optional. There are also some questions meant to help you deepen your understanding of Kotlin and software development as a whole.

I read your PR message, and we'll go through them as we go on with this mission 👍 

I’ll mark this pull request as “changes requested” for now. Once you’ve gone through the review, feel free to request another review.

If you have any questions, feel free to leave a comment in this pull request or DM me anytime.
Happy hacking! 🔥🔥🔥

build.gradle.kts
jieun-pak: Thank you for pointing that out. Since there were no explicit restrictions, we thought it would be fine to add kotlin("test") in the dependencies. 

Thanks to your feedback, I learned two things.
1.  most of the functions provided by kotlin("test") are already available through JUnit. 
2. changing the dependencies in build.gradle.kts affects the project setup, so it should be better to think about it carefully.

I’ll definitely keep it in mind next time.

src/main/kotlin/Application.kt
jieun-pak: Yes, thanks to your feedback and the experience of the 1st week, I learned how to use automatic reformat. I will keep in mind to use it for the next projects. Thank you!

src/main/kotlin/view/InputView.kt
jieun-pak: I thought about my version and suggested version. Since I'm used to C programming, I thought my version has some pros using slightly less memory.

But I ended up thinking that pros of suggested version is more important, since...
1. it has clear separation of tasks
2. it improved readability
3. it's easier to debug

So I would use your version in Kotlin project. Thanks for suggesting me the new version!

src/test/kotlin/model/LottoTest.kt
jieun-pak: I will not forget to use automatic fix! thanks

gradlew.bat
jieun-pak: That file is the Windows batch script which is paired with 'gradlew' and 'gradle/wrapper' directory. I created this file by trying to run it in my personal laptop, and pushed it to this repository. Thanks for pointing it out, I removed it!

src/main/kotlin/view/InputView.kt
jieun-pak: What I did to implement your suggestion was to relocate the ErrorMessages and PromptMessages by creating separate enum classes for them. I'm sure it will make the different types of messages more organized and easier to maintain, instead of having all message constants in 'companion object' in each file.

Thank you for your suggestion!

src/main/kotlin/view/InputView.kt
jieun-pak: I completely understand your concerns about while(true) loops. I've refactored the code to use a more controlled recursive approach with a maximum attempt limit. This approach eliminated the risk of infinite loops while maintaining clear intent. 

This new recursive solution also makes it easier to manage the retry logic and error cases, and it's more idiomatic in functional programming. I appreciate you highlighting this important consideration for writing more maintainable and safer code!

src/main/kotlin/view/InputView.kt
jieun-pak: I fixed it. The input is now trimmed before parsing to Int. Thank you for catching it, I won't miss those basic trimming stuff next time!

src/main/kotlin/view/InputView.kt
jieun-pak: I can't agree more. I created new class 'Ticket' and 'WinningCombination' and replace validate*() functions with the functions of them. My previous code was more like C programming style which I'm used to. Thank you for this comment!

src/main/kotlin/model/Lotto.kt
jieun-pak: I separated it into a dedicated Ticket class as you suggested. Now the list is encapsulated and only handled inside the domain class. It's much safer and easier to test now. Thanks for this feedback!

src/main/kotlin/model/Lotto.kt
jieun-pak: I created a Ticket class for this, and I’m now using it to hold both the user’s tickets and the winning numbers. This change made validation and responsibility separation clearer!

Thanks for the suggestion.

src/test/kotlin/model/LottoTest.kt
jieun-pak: I updated it by creating a 'Ticket' class, and added additional test functions so that we can now manually assign specific numbers for testing. This made testing with exact combinations much easier. Thank you!

src/test/kotlin/model/LottoTest.kt
jieun-pak: I and my peer were not used to this TDD and we were not sure about what were possible in Kotlin. So I must say that adapting to TDD was hard and also not partly possible. But I feel like next time if I put more time planning, I can do TDD much much better.

src/main/kotlin/model/Statistics.kt
jieun-pak: I accept all of your advice. Thus I replaced MutableList with IntArray. As you pointed out, IntArray is more memory-efficient for storing primitive types like integers because it avoids the overhead of "boxing" them into objects.

Also, I used matches.toList(). I learned that this is a great defensive programming practice because it prevents any other part of my code from changing the list of matches. 

Thank you so much!

src/main/kotlin/model/Rank.kt
jieun-pak: Thanks for the hint and the thoughtful question! 
I tried to use Kotlin’s core building blocks.
I already have class, enum class, data class, and object for different purposes. Now I see that Ticket and WinningCombination could work even better as data class since they mostly hold data. And implemented this. Thanks a lot!

jieun-pak:
Hello PK! 🙌

It was awesome to read this high-quality code review! Thank you again for this and thoughtful questions.
What I've learned the most from this code review was about better design of Kotlin. I really appreciate it.

I’ve gone through all the comments and applied the suggested changes.
Please take a look when you have time!🔥

Thanks again for your kind guidance, it really helps. 👍

gradlew.bat
pkeugine: It's okay to keep it in the project.
Usually when working on project as a team, this file would be treated as a separate PR
since it's not related to lotto feature but is still required in the project.

Anyways, nice explanation of what it does 👍

src/main/kotlin/view/InputView.kt
```diff
@@ -0,0 +1,94 @@
...
+class InputView {
+    fun getUserInputAsString(
+        prompt: String,
+        numberOfAttempts: Int = 3,
+    ): String {
+        if (numberOfAttempts <= 0) {
+            throw IllegalArgumentException(ErrorMessages.INPUT_TOO_MANY_ATTEMPT.message) // check
+        }
+        println(prompt)
+        val input = readLine()?.trim()
```
pkeugine, line 15:
```suggestion
        val input = readlnOrNull()?.trim()
```
Let's find out what the differences are between these two.

src/main/kotlin/view/InputView.kt
```diff
@@ -0,0 +1,94 @@
...
+    ): Int {
+        if (numberOfAttempts <= 0) {
+            throw IllegalArgumentException(ErrorMessages.INPUT_TOO_MANY_ATTEMPT.message) // automatically return
+        }
+        // ver1: new
+        val inputAsString = getUserInputAsString(prompt)
+        val input = inputAsString.toIntOrNull()
+        // ver2: original
+        // val input = readLine()?.trim()?.toIntOrNull()
+        return if (input != null) {
```
pkeugine, line 36: I recommend not using comments as notes or keeping code in it since it reduces readability, and it loses it's meaning over time. Comments are very easy to add in between logic, but it's hard to remove them once they get merged into the main branch.

While working on the next step, let's decide which logic to keep and what to remove.

src/main/kotlin/view/InputView.kt
```diff
@@ -0,0 +1,94 @@
+package view
+
+import model.Lotto
+import model.Ticket
+
+class InputView {
+    fun getUserInputAsString(
+        prompt: String,
+        numberOfAttempts: Int = 3,
+    ): String {
```
pkeugine, line 10: If methods are only used internally, let's keep it private.

src/main/kotlin/view/OutputView.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    fun displayMatchResults(results: List<Int>) {
+        println("Winning Statistics")
+        println("------------------")
+        println("3 Matches (5,000 KRW) - ${results[4]} tickets")
+        println("4 Matches (50,000 KRW) - ${results[3]}  tickets")
+        println("5 Matches (1,500,000 KRW) - ${results[2]}  tickets")
+        println("5 Matches + Bonus Ball (30,000,000 KRW) - ${results[1]}  tickets")
+        println("6 Matches (2,000,000,000 KRW) - ${results[0]}  tickets")
+    }
```
pkeugine, line 43: If we're printing multiple lines of string at once, we can make use of [multi-line string](https://kotlinlang.org/docs/java-to-kotlin-idioms-strings.html#use-multiline-strings)!

src/main/kotlin/model/WinningCombination.kt
```diff
@@ -0,0 +1,15 @@
+package model
+
+import view.ErrorMessages
+
+data class WinningCombination(
+    val winningNumbers: Ticket,
+    val bonusNumber: Int,
+) {
```
pkeugine, line 8: Nice work 👍

src/test/kotlin/model/WinningCombinationTest.kt
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("invalidBonusNumbers")
+    fun `should throw exception - create WinningCombination with invalid bonus number`(bonusNumber: Int) {
+        assertThrows<IllegalArgumentException> {
+            val winningTicket = Ticket(listOf(1, 2, 3, 4, 5, 6))
+            WinningCombination(winningTicket, bonusNumber)
+        }
+    }
```
pkeugine, line 47:
```suggestion
    @ParameterizedTest
    @MethodSource("invalidBonusNumbers")
    fun `should throw exception - create WinningCombination with invalid bonus number`(bonusNumber: Int) {
        val winningTicket = Ticket(listOf(1, 2, 3, 4, 5, 6))
        assertThrows<IllegalArgumentException> {
            WinningCombination(winningTicket, bonusNumber)
        }
    }
```
This is a very sophisticated(?) difference, but keeping this idea of **only assert what you really want to assert, keeping the scope as small as possible** is very important. If we put our fixture logic in assertion, test results might show us unpredictable, incorrect results when fixture creates an exception.

src/test/kotlin/model/LottoTest.kt
pkeugine: Nice 👍 
TDD itself is really hard even for me, and when I talk with other developers, many of them also feel that it isn't an easy method.

I think you've done a good job in this step, so I'm sure it'll be better next time!

pkeugine (approved):
Hello Jin~
I'm glad my review helped, and nice job on applying them. It must've been a hard work.

I think we're good for this step so let's move on to the next one!
I left some additional reviews for you to look at,
so please consider them when working on the next step.

If you have any questions or opinions, leave them in the comments or DM me anytime.
As always,
Happy hacking 🚄 🚄 🚄

src/main/kotlin/view/InputView.kt
jieun-pak: You're right, I'll keep in mind!

src/main/kotlin/view/InputView.kt
jieun-pak: Thanks for pointing it out, I learned that readlnOrNull is more modern way/naming of using it!

# readLine()
- Return type: String?
- Since : Kotlin 1.0 (old version)
- When EOF : return `null` -> either handle it or use `!!` to force non-null

# readln()
- Return type: String
- Since: Kotlin 1.6.0
- When EOF: throw `EOFException` (kotlin.io.EOFException), never return null
- Exactly same as: `readlnOrNull()!!` 
-> but using `readlnOrNull()` is more graceful way
-> If you want to crash on EOF, just using `readln()` is enough

# readlnOrNull()
- return type: String?
- since: Kotlin 1.6.0
- When EOF: return `null`

src/main/kotlin/view/InputView.kt
jieun-pak: I'll keep in mind!

jieun-pak:
Thank you so much for your advice, I implemented them all!

PR #4, opened 2025-07-03
[Lotto - Step 1] Submit pebencze's mission

pebencze:
Hello!

Thank you for reviewing my code. Here are some things I would like to get feedback on:

- `LottoHandler`: how is the max-retry-loop? Is it a good idea to have this wrapper class?
- `README`: is it better to work with magic numbers or with placeholders/constants e.g.: 5 vs MAX?
- `LottoMachine`: 
  - is the `fun showChange()` getter a good practice in kotlin?
  - shall I change my `start()` function
- In general: 
  - are my` init{}` blocks too long?
  - are private and public access modifiers used well?
- Tests: would you test more edge cases and/or would you use different methods for testing?
- Calculator:
  - would you rather use `var results: MutableMap<Rank, Int> = null` instead of an empty map?
  - shall the init block rather be a function?


README.md
```diff
@@ -1 +1,69 @@
...
+### Lotto Machine 
+- [x] Purchase Amount 
+  - [x] Change is calculated for the User
+  - [x] Minimum 1_000, (maximum 20_000 KRW)
+- [x] Generates List of Lotto tickets
+  - [x] Each ticket is a random set
+  - [x] Number of tickets equals purchaseAmount / 1000
+
+### Winning Numbers
+- [x] Bonus number in range 0 and 50
```
Wordbe: It feels like a combination of the German and Korean lotteries, doesn't it?
Because the price unit was the 'KRW'.
In Korea, the lottery only has numbers from 1 to 45 😄
Anything is fine — I was just curious!

build.gradle.kts
```diff
@@ -18,6 +18,7 @@ dependencies {
     testImplementation("org.junit.jupiter", "junit-jupiter", "5.10.2")
     testImplementation("org.assertj", "assertj-core", "3.25.3")
     testImplementation("io.kotest", "kotest-runner-junit5", "5.8.0")
+    testImplementation(kotlin("test"))
```
Wordbe: When is kotlin("test") necessary?
From the test code, it seems like JUnit is mainly used, and kotlin("test") and JUnit are mixed together. 

If multiple people are working on this code, what do you think would be the best approach? 🤔

gradlew.bat
```diff
@@ -1,92 +1,92 @@
...
-set EXIT_CODE=%ERRORLEVEL%
-if %EXIT_CODE% equ 0 set EXIT_CODE=1
-if not ""=="%GRADLE_EXIT_CONSOLE%" exit %EXIT_CODE%
-exit /b %EXIT_CODE%
-
-:mainEnd
-if "%OS%"=="Windows_NT" endlocal
-
-:omega
+@rem
```
Wordbe: It doesn't seem like you wrote this gradle.bat file yourself — why was it modified? 🤓

Shall we find out when this file is actually needed?

src/main/kotlin/lotto/Calculator.kt
```diff
@@ -0,0 +1,45 @@
+package lotto
+
+class Calculator(
+    tickets: MutableList<Lotto>,
+    val winningNumbers: WinningNumbers,
+    var results: MutableMap<Rank, Int> = emptyMap<Rank, Int>().toMutableMap(),
```
Wordbe: Your question: 
would you rather use var results: MutableMap<Rank, Int> = null instead of an empty map?

---

I think using emptyMap instead of null has more advantages 😊
It reduces the risk of null pointer exceptions and helps eliminate the need for null checks.

But before using emptyMap, shall we also consider if there are any cases where using null might be better, or any potential downsides of emptyMap?

src/main/kotlin/lotto/Calculator.kt
```diff
@@ -0,0 +1,45 @@
...
+                )
+            results[rank] = results.getOrDefault(rank, 0) + 1
+        }
+    }
+
+    fun calculateReturnRate(purchaseAmount: Int): Float {
+        return (calculateTotalEarnings() / purchaseAmount.toFloat())
+    }
+
+    fun calculateTotalEarnings(): Float {
```
Wordbe, line 31: I recommend exploring the advantages of method chaining and trying to refactor your code accordingly 😉
Kotlin provides a variety of built-in functions to support this style.

src/main/kotlin/lotto/Calculator.kt
```diff
@@ -0,0 +1,45 @@
+package lotto
+
+class Calculator(
+    tickets: MutableList<Lotto>,
+    val winningNumbers: WinningNumbers,
+    var results: MutableMap<Rank, Int> = emptyMap<Rank, Int>().toMutableMap(),
+) {
+    init {
```
Wordbe, line 9: Q: shall the init block rather be a function?

src/main/kotlin/lotto/Calculator.kt
```diff
@@ -0,0 +1,45 @@
...
+        return (calculateTotalEarnings() / purchaseAmount.toFloat())
+    }
+
+    fun calculateTotalEarnings(): Float {
+        var total = 0
+        results.forEach { total += it.key.winningMoney * it.value }
+        return total.toFloat()
+    }
+
+    private fun findMatches(
```
Wordbe, line 35: also here!

src/main/kotlin/lotto/Calculator.kt
```diff
@@ -0,0 +1,45 @@
...
+        var count = 0
+        ticket.numbers.forEach { count += if (winningTicket.numbers.contains(it)) 1 else 0 }
+        return count
+    }
+
+    private fun bonusNumberPresent(
+        ticket: Lotto,
+        bonusNumber: Int,
+    ): Boolean {
+        return ticket.numbers.contains(bonusNumber)
```
Wordbe: How about letting the lotto object take responsibility for checking whether a bonus number exists, instead of the Calculator? 🤔

src/main/kotlin/lotto/InputView.kt
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    fun readBonusNumber(): Int {
+        println(BONUS_NUMBER_MESSAGE)
+        val number = readln().toIntOrNull() ?: throw IllegalArgumentException(ERROR_MESSAGE)
+        println()
+        return number
+    }
+
+    private const val PURCHASE_MESSAGE = "Please enter the purchase amount."
```
Wordbe: Great const message 👏

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -0,0 +1,21 @@
+package lotto
+
+class Lotto(var numbers: List<Int>) {
+    init {
+        require(numbers.size == LOTTO_SIZE) {
```
Wordbe, line 5: Find out what kind of exception the require function throws by default 🤓

And it would be great to include an exception message as well!

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -0,0 +1,21 @@
+package lotto
+
+class Lotto(var numbers: List<Int>) {
```
Wordbe: Why is var necessary here?

src/main/kotlin/lotto/LottoHandler.kt
```diff
@@ -0,0 +1,63 @@
+package lotto
+
+object LottoHandler {
```
Wordbe, line 3: Q: how is the max-retry-loop? Is it a good idea to have this wrapper class?

---

The max-retry-loop seems like a pretty solid strategy, especially from the perspective of preventing user abuse.
It's great that you considered this aspect!

The wrapper class also looks like a good choice.

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,34 @@
+package lotto
+
+class LottoMachine(val purchaseAmount: Int, val tickets: MutableList<Lotto> = emptyList<Lotto>().toMutableList()) {
+    private val numberList = (Lotto.Companion.MIN..Lotto.Companion.MAX)
```
Wordbe: In Kotlin, Companion can be omitted 😊

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,34 @@
...
+    init {
+        require(purchaseAmount in MIN..MAX) {
+            throw IllegalArgumentException()
+        }
+        change = purchaseAmount % TICKET_PRICE
+        ticketCount = (purchaseAmount - change) / TICKET_PRICE
+        generateTickets()
+    }
+
+    fun showChange() = change
```
Wordbe: Q: 
is the fun showChange() getter a good practice in kotlin?

---

In Kotlin, instead of using a show~ style, it's more common to use properties without the get prefix to access getters 🙂

src/main/kotlin/lotto/Application.kt
```diff
@@ -0,0 +1,5 @@
+package lotto
+
+fun main() {
+    LottoHandler.start()
```
Wordbe, line 4: Q:
shall I change my start() function

---

It seems fine to me — but what made you think otherwise?

src/main/kotlin/lotto/OutputView.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    fun displayTotalWinningAmount(totalAmount: Float) {
+        println("Total Earnings $totalAmount KRW.")
+    }
+
+    fun displayWinnings(results: MutableMap<Rank, Int>) {
+        println("Winning Statistics\n------------------")
+        Rank.entries.reversed().forEach {
+            if (it == Rank.MISS) {
+                return@forEach
```
Wordbe: What was return@forEach used for? 🤔

To reduce complexity and improve readability, maybe you could try refactoring it using filter?

src/main/kotlin/lotto/Rank.kt
```diff
@@ -0,0 +1,27 @@
...
+enum class Rank(val countOfMatch: Int, val winningMoney: Int) {
+    FIRST(6, 2_000_000_000),
+    SECOND(5, 30_000_000),
+    THIRD(5, 1_500_000),
+    FOURTH(4, 50_000),
+    FIFTH(3, 5_000),
+    MISS(0, 0),
+    ;
+
+    companion object {
```
Wordbe, line 12: Shall we simplify it using a when (matchCount) expression?  😀

Also, since valueOf is an existing method name, how about changing it to something more appropriate?

Wordbe (changes requested):
Thank you for the thoughtful code and well-prepared questions! 😊

Q: Is it better to work with magic numbers or with placeholders/constants (e.g., 5 vs MAX)?
Using constants makes it easier to understand what the value represents, which is definitely a good practice.

Q: Are my init {} blocks too long?
I think the length is appropriate. If they become longer, feel free to break them into methods.

Q: Are private and public access modifiers used well?
Yes, but it would be even better to make variables private if they're not accessed from outside.

Q: Would you test more edge cases and/or would you use different methods for testing?
More detailed testing is always beneficial. For now, testing all public methods should be enough.

I’ve left responses to the rest of your questions in the comments. Thank you again! 🙌

README.md
pebencze: I actually did not make proper research before deciding on the range. 😆

build.gradle.kts
pebencze: Thank you! Very insightful. I did not realize I have been using kotlin.test, but I have read it actually uses the JUnit API, but it does not support ParameterizedTest, for example. In this case it makes more sense to just go with JUnit and reduce our dependencies. But I think it does not make too much of a difference?

gradlew.bat
pebencze: Good question. It is a windows script that ensures gradle is installed. Since I am not working on a windows machine it is not necessary to have it, but good to have for portability and I think it was changed by intelliJ IDEA or I have accidentially changed something. 
Related question: I still feel a bit lost in the whole build directory structure and I am relying on my IDE. Would you recommend andy resources?

src/main/kotlin/lotto/Calculator.kt
pebencze: We are always assuming a map that contains something, so I think it is fine like this. However, we need to give it a default value.

src/main/kotlin/lotto/Application.kt
pebencze: I was wondering if it was not too long...

src/main/kotlin/lotto/Rank.kt
pebencze: It does not work unfortunately, because we also need to match against hasBonus and that is a Boolean.

build.gradle.kts
Wordbe: It’ll be easier if you stick to one style.
Next time, it might be a good idea to try Kotest instead of JUnit and see which one you find more comfortable 😊

gradlew.bat
Wordbe: If it’s related to Gradle build structure, the official documentation below should be helpful!

https://docs.gradle.org/current/userguide/organizing_gradle_projects.html 😊

src/main/kotlin/lotto/Rank.kt
Wordbe: Ah, you're right. my apologies. I got confused between the parameter and the variable countOfMarch. 😊

src/main/kotlin/lotto/Application.kt
Wordbe: Oh, you were referring to the length of the content inside start.
Yes, if that part gets too long, it would be a good idea to split it up! 😊

Wordbe (approved):
Thank you for reflecting the review feedback.
If possible, I’d love for you to consider the remaining suggestions in the next mission.

Also, I’ve checked that you applied some of my comments through your commit messages, but I’d appreciate it if you could also leave a brief reply on each comment - that would make it easier for me to follow up!

Great work, and I’ll see you in the next mission! 😊

PR #6, opened 2025-07-03
[Lotto - Step1] Submit Apoorva's mission 

Apoorva-S-Natesh:
My team and I tried the TDD approach to navigate the terminal based Lotto program. We wrote tests and then wrote code that passes the tests, then refactored the code and repeat. We tried to follow the Kotlin coding conventions. We think the architecture could be better.
I would like feedback and suggestions on:
1. The overall architecture
2. Coding readability, how to make the code more readable?
3. Alternative methods and approach
4. I am not sure if this is the best way of writing tests

I would really appreciate if you could give me some questions that would help me think and gain a better understanding of Kotlin. Thank you for your time.


README.md
```diff
@@ -1 +1,42 @@
...
+- [x] ticket numbers should not be empty
+
+Winning Numbers
+- [x] winning numbers are in the range 1..45
+- [x] winning numbers not in range 1..45
+- [x] winning numbers are unique 
+- [x] bonus number is in the range 1..45
+- [x] bonus number is not in the range 1..45
+- [x] bonus number in the list of winning numbers
+- [x] bonus number not in the list of winning numbers
```
pkeugine, line 21: Excellent README. Very detailed and well-organized. 💯
Your git commits also look great, with clear names that help readers understand which work has been done.

To take it a step further, it would be even better to divide the work into more focused commits.
For example, commit [test:add bonus number test](https://github.com/herotechcourse/kotlin-lotto/pull/6/commits/d50dd0073c6dfe8394ed760c3b2bae26ae1a15a9) includes not only tests, but also document changes, fixture and naming updates, and bonus number logic.  
While this isn’t wrong, the commit could be split into smaller, more specific commits, or at least have certain parts moved to different commits for clarity.

Personally, when following TDD, I prefer to use `feat:` commits for new logic that includes both the implementation and its corresponding TDD tests or TODOs.
For additional tests or fixtures unrelated to new features, that's when I use `test:` commits instead.
It'll be great for you to talk about this with fellow crews and see how others think about commit messages 😄

build.gradle.kts
```diff
@@ -18,6 +18,7 @@ dependencies {
     testImplementation("org.junit.jupiter", "junit-jupiter", "5.10.2")
     testImplementation("org.assertj", "assertj-core", "3.25.3")
     testImplementation("io.kotest", "kotest-runner-junit5", "5.8.0")
+    testImplementation(kotlin("test"))
```
pkeugine: I see you’ve added a new dependency to the project 😮
Although it’s used in `WinningRankTest`, I think we can achieve the same result using what’s already provided without introducing a new dependency.
**Hint**: [assertJ](https://assertj.github.io/doc/)

Even though it's written in Java, it shouldn’t be too difficult to use,
especially if you check out the documentation I linked above.
It’s actually one of the best resources to start reading when learning programming.

gradlew.bat
```diff
@@ -1,92 +1,92 @@
...
+@rem ##########################################################################
+
+@rem Set local scope for the variables with windows NT shell
+if "%OS%"=="Windows_NT" setlocal
+
+set DIRNAME=%~dp0
+if "%DIRNAME%"=="" set DIRNAME=.
+@rem This is normally unused
+set APP_BASE_NAME=%~n0
+set APP_HOME=%DIRNAME%
```
pkeugine, line 31: This file gets recognized as 'changed' though it doesn't seem like it changed. 🤔 
Can you guess what happened?

Also, what does this `gradlew.bat` file do?

src/main/kotlin/lotto/Const.kt
```diff
@@ -0,0 +1,11 @@
...
+
+object Const {
+    const val PRICE = 1000
+    const val NUMBER_COUNT = 6
+    const val MIN = 1
+    const val MAX = 45
+    const val AMOUNT_ERROR = "Invalid amount, must be divisible by $PRICE"
+    const val DISTINCT_NUM = "Numbers have to be distinct"
+    const val NUM_RANGE = "Numbers should be between 1 to 45"
+}
```
pkeugine: I see you’ve gathered the project’s constants into a single file.
It looks clean at first, but as more classes are added, more and more of them will end up depending on `Const`.
What do you think could be the downside of this?

pkeugine: Additionally, it would be better to use a full, non-abbreviated name instead of `Const`.
Shortened names often come down to personal preference, which can lead to inconsistent naming across the codebase.

src/main/kotlin/lotto/Const.kt
```diff
@@ -0,0 +1,11 @@
+package lotto
+
+object Const {
+    const val PRICE = 1000
+    const val NUMBER_COUNT = 6
+    const val MIN = 1
+    const val MAX = 45
```
pkeugine: I want to know more about `MIN` and `MAX`. What are they min and max of?

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,75 @@
...
+    fun start() {
+        val amount = InputView.inputPurchaseAmount()
+        val ticketList = generateTickets(amount)
+        val winningNumbers = InputView.inputWinningNumbers()
+        val bonusNumber = InputView.inputBonusNumber(winningNumbers)
+        calculateStats(ticketList, winningNumbers, bonusNumber)
+        OutputView.displayResults(winStats, calculateReturnRate(amount))
+    }
+
+    fun generateTickets(amount: Int): List<Ticket> {
```
pkeugine: `generateTickets` method is only used within `LottoMachine`.
Let's keep it private. 🤫 
Please check other methods as well.

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,75 @@
...
+        val ticketCount = amount / Const.PRICE
+        val ticketsList = mutableListOf<Ticket>()
+        repeat(ticketCount) {
```
pkeugine, line 16: Nice use of Kotlin feature `repeat` 👍

src/main/kotlin/lotto/Rank.kt
```diff
@@ -0,0 +1,27 @@
+package lotto
+
+enum class Rank(val countOfMatch: Int, val prize: Int) {
+    FIRST(6, 2_000_000_000),
+    SECOND(5, 30_000_000),
+    THIRD(5, 1_500_000),
+    FOURTH(4, 50_000),
+    FIFTH(3, 5_000),
+    MISS(0, 0),
+    ;
```
pkeugine: Nice use of enum for Rank to represent results. 👍
However, it looks like `countOfMatch` isn’t used anywhere, not even in the `valueOf` method.
Let’s try making use of it, or consider removing it if it feels unnecessary.

src/test/kotlin/lotto/LottoTest.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.junit.jupiter.api.assertThrows
+
+class LottoTest {
+    @Test
+    fun `Amount is greater than 1000`() {
+        assertThrows<IllegalArgumentException> {
+            val lotto = Lotto(2000)
+            requireNotNull(lotto.amount >= 1000)
+        }
+    }
```
pkeugine: This test still passes even if the validation is written incorrectly:
```kotlin
    @Test
    fun `Amount is greater than 1000`() {
            val lotto = Lotto(2000)
            requireNotNull(lotto.amount < 1000)
        }
    }
```
Two things to keep in mind:
* Checking if something is null is different from validating if something is true or false.
* Tests should include some kind of `assertion` in it, rather than `requirements`

With this in mind, let's try rewriting the tests. 🔥 
Please check other tests as well.

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -0,0 +1,8 @@
+package lotto
+
+class Lotto(val amount: Int) {
+    init {
+        require(amount <= Const.PRICE) { "Amount should be greater than or equal to ${Const.PRICE}" }
```
pkeugine: This `require` logic seems to be requiring the opposite of what the message is saying. 🤔
Might be worth double checking the condition and the message to make sure they match.

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -0,0 +1,8 @@
+package lotto
+
+class Lotto(val amount: Int) {
```
pkeugine: This `Lotto` class doesn’t seem to be used anywhere in the application logic.
Could it be a leftover from refactoring? If so, let's remove it.

src/main/kotlin/lotto/OutputView.kt
```diff
@@ -0,0 +1,36 @@
...
+------------------
+3 Matches (${formatNumber(Rank.FIFTH.prize)} KRW) - ${winStats[Rank.FIFTH]} tickets
+4 Matches (${formatNumber(Rank.FOURTH.prize)} KRW) - ${winStats[Rank.FOURTH]} tickets
+5 Matches (${formatNumber(Rank.THIRD.prize)} KRW) - ${winStats[Rank.THIRD]} tickets
+5 Matches + Bonus Ball (${formatNumber(Rank.SECOND.prize)} KRW) - ${winStats[Rank.SECOND]} tickets
+6 Matches (${formatNumber(Rank.FIRST.prize)} KRW) - ${winStats[Rank.FIRST]} tickets
+Total return rate is $returnRate (A rate below 1 means a loss)
+""".trimIndent()
+        println(text)
+    }
```
pkeugine: Nice use of a multi-line string 👍
Is the string unindented for output formatting reasons?

If so, try indenting it like the rest of your logic and see how the output looks 😄 :
```suggestion
    fun displayResults(
        winStats: Map<Rank, Int>,
        returnRate: String,
    ) {
        val text = """
        Winning Statistics
        ------------------
        3 Matches (${formatNumber(Rank.FIFTH.prize)} KRW) - ${winStats[Rank.FIFTH]} tickets
        4 Matches (${formatNumber(Rank.FOURTH.prize)} KRW) - ${winStats[Rank.FOURTH]} tickets
        5 Matches (${formatNumber(Rank.THIRD.prize)} KRW) - ${winStats[Rank.THIRD]} tickets
        5 Matches + Bonus Ball (${formatNumber(Rank.SECOND.prize)} KRW) - ${winStats[Rank.SECOND]} tickets
        6 Matches (${formatNumber(Rank.FIRST.prize)} KRW) - ${winStats[Rank.FIRST]} tickets

        Total return rate is $returnRate (A rate below 1 means a loss)
    """.trimIndent()

        println(text)
    }
```

It would be great if you could look into what allows our multi-line string to maintain its formatting in the console, even when we follow proper indentation in the code.

src/main/kotlin/lotto/Ticket.kt
```diff
@@ -0,0 +1,8 @@
+package lotto
+
+class Ticket(val numbers: List<Int>) {
+    init {
+        require(numbers.size == 6 && numbers.distinct().size == 6) { "Numbers have to be distinct" }
+        require(numbers.all { it in 1..45 }) { "Numbers should be between 1 to 45" }
+    }
```
pkeugine: Let's try creating a class for numbers in `Ticket` instead of using `Int` directly.
By doing so, we'll be able to make it responsible for checking number range. 👍

pkeugine: Also, I remember seeing `NUMBER_COUNT`, `MIN`, and `MAX` in `Const` class.
It'll be great if we could make use of them here.

src/main/kotlin/lotto/OutputView.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    private fun formatNumber(prizeAmount: Int): String {
+        return prizeAmount.toString().reversed().chunked(3).joinToString(",").reversed()
+    }
```
pkeugine: Nice job formatting the prize amount 👍
As we continue learning Kotlin (and the underlying Java), 
we’ll be able to make even better use of the tools available to us. 😄
```suggestion
    private fun formatNumber(prizeAmount: Int): String {
        return NumberFormat.getNumberInstance().format(prizeAmount)
    }
```
I recommend exploring how these built-in classes work under the hood.
It’s a great way to deepen your understanding.

**Tip**: Using `Cmd + B` (or `Ctrl + B` on Windows) with your cursor on a class or method will take you to its source code.
It’s super helpful for understanding how things work. I use it all the time. :)

src/main/kotlin/lotto/WinningTicket.kt
```diff
@@ -0,0 +1,11 @@
...
+
+class WinningTicket(val winningNumbers: List<Int>, val bonusNumber: Int = 1) {
+    init {
+        require(winningNumbers.size == 6) { "There should be 6 winning numbers" }
+        require(winningNumbers.all { it in 1..45 }) { " Winning numbers should be between 1 to 45 " }
+        require(winningNumbers.toSet().size == 6) { " Winning numbers should be unique " }
+        require(bonusNumber in 1..45) { "Bonus number should be between 1 to 45" }
+        require(bonusNumber !in winningNumbers) { "Bonus number should be different from winning numbers" }
+    }
+}
```
pkeugine: The previous review about separating classes and using named constants also applies to `WinningTicket`.
Additionally, since `WinningTicket` shares the same validation logic for the numbers, could we reuse the `Ticket` class instead?

**Hint**: composition

pkeugine (changes requested):
Hello Apoorva 🙌
Congrats on your very first PR here! 😊
I see you’ve done a great job with class separation and made good use of various Kotlin features.

I’ve read your PR message, and we’ll go over it through some reviews and questions as we dive into the mission.

I’ve added comments on how we could improve the logic, along with a few questions to get you thinking with some tips! Hope they help 😄

I’ll mark this PR as "changes requested" for now.
Once you’ve gone through the reviews and made updates, feel free to send another review request and we’ll go over the changes together.

If you have any questions, feel free to leave a comment on this PR or DM me anytime.
Happy hacking! 💪💪💪

README.md
Apoorva-S-Natesh: Noted!
The coaches also mentioned that this is a good practice! when the logic and the test are implemented then its a `feat:`.
Is TCR (Test && Commit || Revert) a good practice while following TDD or does it create a confusing git log and commit history?

gradlew.bat
Apoorva-S-Natesh: This happened as the .gitattributes file was changed. On windows the Line Endings are CRLF and on macOS its LF. Git auto-converts line endings if config has core.autocrlf=true or similar
If gradlew.bat was committed with one type of line ending and then saved with another. Git thinks it has changed, even if the file's content appears the same.
the .gitattributes file was changed it caused issue to commit the file.
 
gradlew.bat is a Windows batch script that is part of the Gradle Wrapper. It allows to execute Gradle tasks without requiring a pre-installed version of Gradle on the system.

src/main/kotlin/lotto/Const.kt
Apoorva-S-Natesh: I agree, the downside would be that a single giant Constants object quickly becomes bloated and hard to navigate. Refactoring like renaming and tracking usages becomes harder, and code becomes tightly coupled to the constants object.

src/main/kotlin/lotto/Rank.kt
Apoorva-S-Natesh: countOfMatch  is not used anywhere and feels unnecessary

src/main/kotlin/lotto/OutputView.kt
Apoorva-S-Natesh: wow! This is super helpful! Thank you

src/main/kotlin/lotto/WinningTicket.kt
Apoorva-S-Natesh: Thank you for the suggestion to refactor. I wanted to confirm my understanding of when to use inheritance and when to use composition.
Inheritance when we have a clearly defined class hierarchy, where one class represents a more general concept and another class represents a more specific concept. when the relationship is "is-a".
Composition when we want to combine objects and change their behavior without impacting other parts of the code. when the relationship is "has-a". Like the WinningTicket "has-a" Ticket(numbers) and bonusNumber.

src/main/kotlin/lotto/Ticket.kt
Apoorva-S-Natesh: Is it better to declare the MINIMUM_NUMBER and MAXIMUM_NUMBER number of LottoNumber inside a companion object in order to access it outside of the class or it is better to make it private?

README.md
pkeugine: I think it depends on the workflow of the team or the scale of the work the developer is working on.
The best way to find out if it is a good practice or not is to try it and see if it works well in your environment.

As for the git log and commit history, I tend to edit and rebase often since there's always unexpected changes even after I commit my works. It might take some time to get used to rebasing at first, but once you get the hang of it, it's a very strong option when organizing git history.

src/main/kotlin/lotto/Const.kt
pkeugine: Perfect 💯 
This also causes frequent git conflict since multiple developers will try to edit this one single `Const` file all at once.

src/main/kotlin/lotto/Ticket.kt
pkeugine: I would try to leave it private, but there could be some cases where it really needs to be used in some place else.
If I feel like that logic belongs somewhere else rather than `LottoNumber`, then that's when I would change the access level to public.

src/main/kotlin/lotto/WinningTicket.kt
pkeugine: Yes, I think your understanding is spot on.
Nice work summarizing the knowledge 👍

src/test/kotlin/lotto/WinningRankTest.kt
```diff
@@ -0,0 +1,36 @@
+package lotto
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class WinningRankTest {
+    @Test
+    fun `return FIRST when 6 numbers match`() {
+        val rank = Rank.valueOf(6, false)
+        assertThat(rank).isEqualTo(Rank.FIRST)
```
pkeugine, line 10: perfect use of assertJ 💯

src/main/kotlin/lotto/InputView.kt
```diff
@@ -0,0 +1,50 @@
+package lotto
+
+object InputView {
+    fun inputPurchaseAmount(): Int {
+        while (true) {
+            try {
+                println("Please enter the purchase amount.")
+                val amount = readln().trim().toInt()
+                AmountValidator.validate(amount)
```
pkeugine, line 9: I think it would be better to move cost(`amount`) validation to `Ticket` class rather than `InputView, since `InputView`'s responsibility is to obtain users' input, and validating the amount is more of a 'business logic'.

If I were to join this team and maintain the project, I wouldn't search `InputView` for any domain related validation.

src/main/kotlin/lotto/WinningTicket.kt
```diff
@@ -0,0 +1,10 @@
+package lotto
+
+class WinningTicket(val winningNumbers: Ticket, val bonusNumber: LottoNumber) {
+    init {
+        require(bonusNumber.toInt() in Constants.MINIMUM_NUMBER..Constants.MAXIMUM_NUMBER) {
+            "Bonus number should be between ${Constants.MINIMUM_NUMBER}..${Constants.MAXIMUM_NUMBER}"
+        }
+        require(bonusNumber !in winningNumbers.numbers) { "Bonus number should be different from winning numbers" }
+    }
```
pkeugine, line 9: composition 💯 💯 💯

src/main/kotlin/lotto/LottoNumber.kt
```diff
@@ -0,0 +1,26 @@
+package lotto
+
+class LottoNumber private constructor(private val value: Int) {
+    companion object {
+        private const val MINIMUM_NUMBER = 1
+        private const val MAXIMUM_NUMBER = 45
+        private val NUMBERS: Map<Int, LottoNumber> = (MINIMUM_NUMBER..MAXIMUM_NUMBER).associateWith(::LottoNumber)
```
pkeugine, line 7: This looks interesting 😮 
I'm curious about the thought behind this way of keeping the lotto numbers.

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,75 @@
...
+            val ticket = Ticket(generateTicketNumbers())
+            ticketsList.add(ticket)
+        }
+        OutputView.displayTickets(ticketCount, ticketsList)
+        return ticketsList
+    }
+
+    private fun generateTicketNumbers(): List<LottoNumber> {
+        return (LottoNumber.allNumbers()).shuffled().take(Constants.NUMBER_COUNT).sortedBy { it.toInt() }
```
pkeugine, line 25: Since we have LottoNumber that holds all the valid numbers in it,
rather that taking them out with `allNumbers`,
we could let it handle this and just call `LottoNumber.takeRandom(NUMBER_COUNT)`.

pkeugine (approved):
Hello Apoorva,
nice work refactoring the project!

I've added some additional comments.
Let's see if we can consider refactoring those parts as we work on the next step.
There's also some questions so feel free to answer them in this PR or leave it in the next PR message.

Nice work and I'll see you again soon!
Happy hacking 🥇 🥇 🥇

src/main/kotlin/lotto/Ticket.kt
Apoorva-S-Natesh: Understood :) Thank you

src/main/kotlin/lotto/InputView.kt
Apoorva-S-Natesh: Understood, are there good reasons to re-validate everything along the chain or will there be overheads?

src/main/kotlin/lotto/LottoMachine.kt
Apoorva-S-Natesh: This is more straight forward! Thank you, will implement this

src/main/kotlin/lotto/LottoNumber.kt
Apoorva-S-Natesh: The LottoNumber in kept this way to implement SRP (single responsibility Principle from SOLID) which was introduced in the instructor led session. To separate the numbers and wrap the primitive Int and to test them separately.

src/main/kotlin/lotto/InputView.kt
Apoorva-S-Natesh: Implemented

PR #7, opened 2025-07-03
[Lotto - step1] Submit SebasNadu's mission

SebasNadu:
## Lotto Step 1
This PR show the implementation of lotto that allows users to purchase tickets, enter winning numbers, and calculate their winnings based on matching numbers. 

#### Domain
- Lotto class with the logic of correct lotto ticket
- WinningCombination Class contains the winning numbers and bonus number of the week
- Rank Enum with the rank of prizes in relation of the matched numbers

#### Controller
- Flow of the application
#### Services
- LottoService extract logic from the controller to improve testability and extendability
#### Utils
- LottoUtils contains helper functions for the app
#### View
- InputView and OutputView
#### Exceptions
- LottoExceptions customize the exceptions for error cases during the runtime of the app.
#### Unit tests
- Implemented TDD for the development of the app.

### Feedback Welcome
We’d appreciate your thoughts on the following:

- Project structure: Is the current package and file organization appropriate for the app’s scope?
- Testing depth: Are we covering enough edge cases, or should more scenarios be added?
- Documentation: Would more comments or README details improve maintainability?
- Kotlin practices: Are we aligning well with Kotlin idioms and style conventions?

The implementation is ready for review. Your feedback is very welcome and appreciated.

Thanks for reviewing!
— Sebastian Navarro

TODO:
- E2E test.


README.md
```diff
@@ -1 +1,63 @@
...
+### Lotto Class
+
+- [x] 6 unique numbers
+- [x] in range 1 - 45
+- [x] Lotto instance has toString
+
+### WinningNumbers Class
+
+- [x] valid lotto tickets inside winning numbers class
+- [x] valid 1 bonus unique number inside winning numbers class
```
pkeugine: Nice job organizing the README document 👍

I also noticed you’ve stacked quite a few commits for this project by separating TDD related tests, implementation, and documentation into their own commit. 😮
How was that experience? Did it feel difficult? I’d love to hear how TDD and working with small, focused commits felt for you.

Some developers follow the principle that 
_“Every commit should be in a finished state, meaning all tests pass and the application builds successfully.”_
Do you agree with this idea? If so, how might we change the way we stack our commits to follow that approach?

README.md
```diff
@@ -1 +1,63 @@
...
+- [x] customize exception for DRY
+- [x] add documentation
+- [x] separate controller logics
+- [x] refactor methods
+
+---
+
+### TODOS
+
+- [ ] add **E2E test**
```
pkeugine: Oh man, 🔥 E2E test 🔥 is on it's way huh 😄 
How did this idea come up to you and your pair?
And what IS an E2E test anyway? 🤔

src/main/kotlin/lotto/domain/Lotto.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import lotto.exceptions.LottoException
+import lotto.util.LottoUtils
+
+/***
+ * This Class represents a lotto ticket,
+ *  - contains 6 unique numbers
+ *  - in range 1 to 45
+ */
+class Lotto(val numbers: List<Int> = emptyList<Int>()) {
```
pkeugine:
```suggestion
class Lotto(val numbers: List<Int> = emptyList()) {
```
You can remove explicit type argument here.
But how is that possible? It would be great if you could look into this and tell me about it :)

**Hint**: type inference, generic function

src/test/kotlin/lotto/domain/LottoTest.kt
```diff
@@ -0,0 +1,48 @@
...
+        assertThrows<IllegalArgumentException> {
+            Lotto()
+        }
+    }
+
+    @Test
+    fun `Lotto has 6 numbers`() {
+        val ticket = listOf(1, 2, 3, 4, 5, 6)
+        assertThat(Lotto(ticket))
+    }
```
pkeugine: Nice start on testing the logic!
However, this test currently isn’t asserting anything.
Try adding a period(`.`) after `assertThat()` and you’ll see a whole list of assertion methods provided by AssertJ:
image

Let’s make sure our tests actually assert the behavior we want to verify.
If you need a starting point, here’s the [official AssertJ documentation.](https://assertj.github.io/doc/) 
It’s beginner-friendly and I highly recommend it. 👍 

--- 

I see that you've used some of these on other tests. Let's try out other methods as well.

src/main/kotlin/lotto/view/InputView.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    fun getWinningNumbers(): List<Int> {
+        println(Constants.WINNING_NUMBERS_PROMPT)
+        val input = readln()
+        return input
+            .split(',')
+            .map {
+                it.trim().toIntOrNull()
+                    ?: throw LottoException.InvalidWinningNumbersFormatException(input)
+            }
```
pkeugine: Nice job trimming the user input for purchase amount 👍 
Why don't we apply that to winning numbers as well?

src/main/kotlin/lotto/view/InputView.kt
```diff
@@ -0,0 +1,34 @@
...
+        println(Constants.BONUS_NUMBER_PROMPT)
+        val input = readln()
+        return input.trim().toIntOrNull() ?: throw LottoException.InvalidBonusNumberFormatException(input)
+    }
+
+    object Constants {
+        const val PURCHASE_PROMPT = "Please enter the purchase amount."
+        const val WINNING_NUMBERS_PROMPT = "Please enter last week’s winning numbers."
+        const val BONUS_NUMBER_PROMPT = "Please enter the bonus number."
+    }
```
pkeugine: I see there’s an object nested inside another object.
It looks like you may have followed the pattern used for companion objects. (let me know if I misunderstood!)

That said, I don’t think we need to keep these constants in a separate `Constants` object.
How could we simplify this?

**Hint**: Keeping things simple is often a great choice. 😄

Also, what are the characteristics of `companion object`s?

src/main/kotlin/lotto/controller/LottoController.kt
```diff
@@ -0,0 +1,63 @@
...
+            ranks.add(result)
+        }
+        return ranks
+    }
+
+    private fun getTickets(): Pair<Int, List<Lotto>> {
+        val amount =
+            LottoUtils.retryUntilSuccess({
+                InputView.getPurchaseAmount().also { LottoService.purchaseAmountValidator(it) }
+            })
```
pkeugine:
```suggestion
            LottoUtils.retryUntilSuccess {
                InputView.getPurchaseAmount().also { LottoService.purchaseAmountValidator(it) }
            }
```

Lambda expression can be simplified :)

src/main/kotlin/lotto/util/LottoUtils.kt
```diff
@@ -0,0 +1,39 @@
...
+     */
+    fun <T> retryUntilSuccess(block: () -> T): T {
+        while (true) {
+            try {
+                return block()
+            } catch (e: LottoException) {
+                println(e.message)
+            }
+        }
+    }
```
pkeugine: I see you've used some dark magic (generics) here, except... they're not dark and also not magic I guess.
Nice work 👍 

However, two things came up to my mind when I saw this logic:

1. By the looks of where this method is used, I think this logic belongs in `InputView` rather than **Lotto**Utils. Lotto application doesn't need to know whether the input logic retries each time the input is wrong. They just want the final valid input, and they just go with it.

2. `while (true)` is a very dangerous move when it comes to preventing unexpected bugs. Yes, it may look unharmful for now, and I agree that it's very straight forward when it comes to retrying something until the expected result comes out. However, when maintaining the project, anything can change, and if that change doesn't break out of this never ending while loop, bam, our application changes from lotto application to never-ending-keyboard-input-practice-application.

So why don’t we give refactoring this a try? 🙂 
I'm open to ideas.

src/main/kotlin/lotto/services/LottoService.kt
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    fun generateLottoTickets(amount: Int): List<Lotto> {
+        val numberTickets = amount / Lotto.Companion.PRICE_OF_TICKET
+        return List(numberTickets) { getLottoTicket() }
+    }
+
+    fun getRankFrom(
+        ticket: Lotto,
+        winningNumbers: WinningCombination,
```
pkeugine: Nice job creating classes and separating responsibilities! 👍 
To be more clear, why don't we match the parameter name `ticket` with it's type `Lotto`?
Since `ticket` and `lotto` get used a lot in lotto domain, having names fixed helps teammates understand the logic more easily.

src/main/kotlin/lotto/domain/WinningCombination.kt
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * the class contain winning ticket and bonus number to eval user's Lotto tickets
+ */
+class WinningCombination(val winningTicket: Lotto, val bonusNumber: Int = 0) {
+    init {
+        LottoUtils.requireOrThrow(
+            LottoUtils.isInRange(bonusNumber),
+            LottoException.InvalidBonusNumberException(bonusNumber),
+        )
```
pkeugine:
```suggestion
        isInRange(bonusNumber).orThrow { LottoException.InvalidBonusNumberException(bonusNumber) }
```
One of Kotlin’s strengths is its ability to build DSLs and extend existing language features in a natural way.
With that in mind, we can actually refactor this logic to make it more expressive like the example above.

Let’s dive into Kotlin a bit and explore how we can achieve this!😄 
**Hint**: [extension function](https://kotlinlang.org/docs/extensions.html#extensions-are-resolved-statically)

src/main/kotlin/lotto/domain/Lotto.kt
pkeugine: Describing how the class works with range information really helps. Nice job. 👍

src/main/kotlin/lotto/domain/Lotto.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    override fun toString(): String {
+        return numbers.joinToString(", ", "[", "]")
+    }
+
+    companion object {
+        const val LOTTO_PICK_SIZE = 6
+        const val MIN_RANGE = 1
+        const val MAX_RANGE = 45
+        const val PRICE_OF_TICKET = 1000
```
pkeugine:
```suggestion
        const val PRICE_OF_TICKET = 1_000
```
Just like how you expressed other numbers in the project, let's try to keep that convention.

src/main/kotlin/lotto/services/LottoService.kt
```diff
@@ -0,0 +1,51 @@
...
+    private fun getLottoTicket(): Lotto {
+        val numbers = generateRandomNumbers(Lotto.Companion.LOTTO_PICK_SIZE)
+        return Lotto(numbers)
+    }
+
+    private fun generateRandomNumbers(size: Int): List<Int> {
+        return (Lotto.MIN_RANGE..Lotto.MAX_RANGE)
+            .shuffled()
+            .take(size)
+    }
```
pkeugine:
```suggestion
    private fun getLottoTicket(): Lotto {
        val numbers = generateRandomNumbers()
        return Lotto(numbers)
    }

    private fun generateRandomNumbers(): List<Int> {
        return (Lotto.MIN_RANGE..Lotto.MAX_RANGE)
            .shuffled()
            .take(Lotto.LOTTO_PICK_SIZE)
    }
```

If the method is used only once and the parameter can be removed, I think it's good to simplify the logic.
When we need to use it somewhere else, that's when we can start refactoring.
I'm curious about your thoughts on this :)

Also, `Companion` can be removed when using constants so please check this in other places as well.

src/main/kotlin/lotto/exceptions/LottoException.kt
```diff
@@ -0,0 +1,34 @@
+package lotto.exceptions
+
+import lotto.domain.Lotto
+
+sealed class LottoException(msg: String) : IllegalArgumentException("[ERROR]: $msg") {
```
pkeugine, line 7: I see you've used sealed class, object, class, and enum for this project 😮 
What made you use `sealed class` for organizing exceptions?
In Kotlin, we can actually organize multiple classes in single file without sealed classes as well.
I'm not saying this is wrong. I just want your thoughts on this :)

src/main/kotlin/lotto/domain/WinningCombination.kt
```diff
@@ -0,0 +1,20 @@
...
+        LottoUtils.requireOrThrow(
+            winningTicket.numbers.none { it == bonusNumber },
+            LottoException.InvalidBonusNumberException(bonusNumber),
+        )
+    }
+}
```
pkeugine: Nice job using `Lotto` in `WinningCombination` 👍 
By doing this, `WinningCombination` gets all the `Lotto` requirements as well!

src/test/kotlin/lotto/services/LottoServiceTest.kt
```diff
@@ -0,0 +1,36 @@
...
+        }
+    }
+
+    @ValueSource(ints = [TICKET_PRICE - 1, TICKET_PRICE + 1, TICKET_PRICE * 2 + 1])
+    @ParameterizedTest
+    fun `throw if ticket price is invalid`(ticketPrice: Int) {
+        assertThrows<IllegalArgumentException> {
+            LottoService.purchaseAmountValidator(ticketPrice)
+        }
+    }
```
pkeugine: Nice use of `@ValueSource` & `@ParameterizedTest` 👍 
Test names are also readable.
To go beyond, I recommend making test name conventions with your pair,
since having formatted, consistent test names really help the team maintain them,
especially when the numbers of tests go up.

src/test/kotlin/lotto/domain/LottoTest.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    @Test
+    fun `Lotto numbers are not in range`() {
+        val ticket = listOf(1, 100, 2, 3, 4, 5)
+        assertThrows<IllegalArgumentException> {
+            Lotto(ticket)
+        }
+    }
```
pkeugine: The second test about lotto numbers not being in range could be improved by using boundary values such as `0` and `46`.

src/main/kotlin/lotto/services/LottoService.kt
```diff
@@ -0,0 +1,51 @@
...
+object LottoService {
+    private const val ZERO_DOUBLE = 0.0
+
+    fun purchaseAmountValidator(amount: Int) {
+        if (amount < Lotto.Companion.PRICE_OF_TICKET || amount % Lotto.Companion.PRICE_OF_TICKET != 0) {
+            throw LottoException.InvalidAmountException(
+                amount,
+            )
+        }
+    }
```
pkeugine: I personally try to name methods in `verb`s. Sometimes this doesn't work, but in this case, I think we can make it possible. Why do you think I try to name the methods as verbs? How should we name classes then? Variables? Please tell me your thoughts on this, I'm curious :)

pkeugine: Also, since this validation uses `PRICE_OF_TICKET` in `Lotto` class, I think `Lotto` class is responsible for holding this logic.

src/main/kotlin/lotto/domain/Lotto.kt
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    override fun toString(): String {
+        return numbers.joinToString(", ", "[", "]")
+    }
+
+    companion object {
+        const val LOTTO_PICK_SIZE = 6
+        const val MIN_RANGE = 1
+        const val MAX_RANGE = 45
```
pkeugine: Let's say `Lotto` is responsible for holding 6 lotto numbers and having logics that are related to those six numbers.
Then I think there should be a separate `LottoNumber` class that validates if the number is actually valid or not.
How does this sound?

pkeugine (changes requested):
Hello Sebas, congrats on your first PR here! 🥳 
I see you've done a great job separating responsibilities in various classes.

I left some reviews along with suggestions, and (quite a lot of) questions that I think will help you go deep into Kotlin... or just because I was curious. 😂 

I've read your PR message, and I think our comments will be the answers to them.
I'll keep this PR as a 'request changes' for now, so when you finish going over the comments, go ahead and ask for another review.

If you have any questions, just hit me with it via PR comments or DM.
Happy hacking! 🚀 🚀 🚀

README.md
SebasNadu: Yes, It was a bit frustrating at first because I usually work by implementing everything first and then committing the completed feature. In the beginning, this step-by-step commit approach sometimes breaks my flow. For example, after writing the test, I often want to jump straight into implementing the logic.

But, I understand the value of this workflow. If something goes wrong, it's easier to go back to a specific commit without wasting time. It's also better for debugging and helps others follow the implementation. Reviewing small, focused commits is much clearer than dealing with a large commit containing many unrelated changes.

So, I might try writing smaller tests to keep the commits clean. What do you think about my thoughts? I’m also curious, what’s your usual workflow in your projects?

README.md
SebasNadu: In the pre-course, we had a small E2E test, and I found it helpful for testing the entire application flow. An E2E test simulates the full user interaction, from input to output, and verifies that the app behaves as expected from a user's perspective.

I noticed that the pre-course used the nextstep testing library to support this, which made things easier. Implementing such a test ourselves would be more challenging, since we'd have to manually capture and simulate stdin and stdout using java.

For step 2, I think it makes more sense to focus on unit tests, especially for classes like ViewInput and ViewOutput. We’ve left a TODO for E2E testing as a possible future improvement. For now, I believe the unit test coverage is sufficient.

What do you think?

src/main/kotlin/lotto/domain/Lotto.kt
SebasNadu: Yes, totally. Kotlin infer the type, emptyList() is a generic function that infer the type depending the context. So yes your are right, that's very redundant.

src/test/kotlin/lotto/domain/LottoTest.kt
SebasNadu: Sorry about that🙏, that was an oversight on our part. I’ll double-check the code to avoid similar mistakes in the future and will review and rewrite the tests as needed.

src/main/kotlin/lotto/view/InputView.kt
SebasNadu: I’m trimming each individual number inside map, so leading/trailing spaces around commas are already handled. That’s why I didn’t apply trim() to the whole input string, but I see how doing that too could improve readability or consistency with the other functions. Let me know if you'd prefer that style.

src/main/kotlin/lotto/view/InputView.kt
SebasNadu: You're right, it's not strictly necessary. I used the separate Constants object to improve readability and structure, but I agree that if const val is written in uppercase, it’s already clear by convention that it’s a constant. I’ll simplify it as you suggested, thanks for the reminder.

Regarding companion objects: they’re special objects inside a class that are compiled as static members. They're useful when you want to define constants or factory methods tied to the class itself, without needing an instance.

src/main/kotlin/lotto/controller/LottoController.kt
SebasNadu: ohhh cool, thank you very much!

src/main/kotlin/lotto/util/LottoUtils.kt
SebasNadu: Thanks for the feedback, I appreciate your thoughts!

About the first point: I see where you're coming from, but I intentionally kept this logic out of InputView because I see its role strictly as handling raw input, not validating it or managing control flow. For me, the point where the view and model connect, where validation and instance creation happen, belongs in the controller. That’s where I feel retryUntilSuccess fits more naturally, and more if I leave some validation inside the model classes like you suggest later.

I placed it in LottoUtils because it’s a reusable template with no business logic, just a control mechanism to retry a lambda until success. Since it’s generic and doesn’t have any Lotto-specific behavior, keeping it in a utility file felt cleaner.

Regarding the while (true), good point. It’s definitely something that needs to be handled with care. One idea would be to limit the number of retries or allow an exit strategy for future flexibility. But for now, since it's only used for user input in a controlled context, I kept it simple. I'm open to adjusting it if we see potential risks down the line.

Happy to hear your thoughts on where we might draw the line between view responsibility and input control flow, Do you think that for the step2 limit a number of tries in the input is a good idea, because it is out from the pre-course requirements, What do you think about that? 🙂

src/main/kotlin/lotto/services/LottoService.kt
SebasNadu: Yes, I totally agree! I'll modify it and next time I’ll pay more attention to these details.

src/main/kotlin/lotto/domain/WinningCombination.kt
SebasNadu: That's a great idea! I have to admit the concept of DSLs is still quite new to me, but I can already see how powerful it is.
It reminds me of operator overloading in C++, also powerful, but dangerous if misused. Without a clear purpose, it can add unnecessary complexity and hurt maintainability.

I'm especially still getting used to lambdas with receivers, they feel a bit tricky to me. Do you have any tips or mental models for understanding them better?

Back to the code, I agree, using an extension function here makes it much more expressive. What do you think about this version?
``` kotlin
fun Boolean.orThrow(lottoException: () -> LottoException) {
    if (!this) throw lottoException()
}

```
It feels clean and reusable.
Curious to hear your thoughts!👀

src/main/kotlin/lotto/services/LottoService.kt
SebasNadu: You're absolutely right, simplifying generateRandomNumbers() by removing the unused parameter makes the code cleaner, especially since it's only used in one place. I don't know what I was thinking😔.

Good point about the Companion😩. I’ll pay more attention to those details, especially when I’m not the one typing during pair programming. It’s easy to overlook small things like that in the moment. Thanks for pointing it out!

src/main/kotlin/lotto/exceptions/LottoException.kt
SebasNadu: Yes, it’s definitely possible to declare multiple classes inheriting from the same superclass in one file. However, using a sealed class enforces that all subclasses are declared in that same file, preventing unexpected extensions from outside. This helps me keep all LottoException types centralized and controlled,  which improves code maintainability and clarity.

Plus with sealed class the IDE give me a nice autocomplete 😅.

That said, I’m open to changing it if it feels like overengineering in this context. What’s your usual approach when organizing custom exceptions?

src/test/kotlin/lotto/services/LottoServiceTest.kt
SebasNadu: Thank you! I appreciate the feedback, it's a very good idea establish some naming conventions 💪

src/main/kotlin/lotto/services/LottoService.kt
SebasNadu: Yes, I totally agree. Method or Function names as verbs make code easier to read like natural language, because they describe actions the program performs. For example, validatePurchaseAmount() clearly says what it does.

Classes should be named as nouns, since they usually represent entities or concepts for example Lotto or Rank.

Variables should describe what they hold, so concise nouns or noun phrases make sense.

About the `PRICE_OF_TICKET`, you're right again, other reason to move this logic to the model 🤔.

Thanks for raising both points!

src/main/kotlin/lotto/domain/Lotto.kt
SebasNadu: Yes 😊 We actually started discussing this in our group, the idea of wrapping all primitives and applying the First-Class Collection principle.

At first, I have to admit it felt like overengineering. Wrapping a single integer in a class? Creating an entire object just for one primitive, only for the sake of readability and maybe maintainability? It seemed excessive.

But then I came across some helpful tricks like @JvmInline value classes, which reduce the overhead. And Jason introduced us to a neat pattern using the companion object as a kind of factory, reusing a static collection of pre-instantiated objects. That was very interesting to me.

Now I’m curious, what downsides or limitations do you see in applying this pattern specifically to something like LottoNumber?

SebasNadu:
Thanks to the feedbacks I addressed the following:

- Fixed issues related to implicit conversions and incorrect test assertions.
- Introduced a LottoNumber class to encapsulate validation logic for individual numbers.
- Replaced List with Set in the Lotto class to enforce uniqueness.
- Removed the unused services layer. Much of the logic was better suited to the domain layer, so I created the LottoMachine and ResultAnalyzer classes and moved relevant logic there from the controller and the services.
- Improved test naming using a consistent convention and added additional tests for better coverage, including edge cases.
- Removed unused utility functions and moved retryUntilSuccess to the controller, since it’s only used there.
- Simplified InputView and OutputView by moving their constants outside the objects for better readability.

Overall, I tried to apply the feedback thoughtfully and improve the structure, readability, and maintainability of the code.

README.md
pkeugine: I prefer having implementation, documents, and tests in a same commit.
Your point on this topic is also relatable, and this is just the first step out of many,
so I'm sure you'll be able to find the way that fits you and your (future) team.

README.md
pkeugine: Great. It'll be nice if you could also look into the difference between integration tests and E2E tests,
since you'll be writing those tests in the near future. :)

src/main/kotlin/lotto/view/InputView.kt
pkeugine: Oops, this is my bad.
When reviewing, I completely overlooked the `trim()` in map.
Thanks for reminding me, and let's skip this review :)

src/main/kotlin/lotto/util/LottoUtils.kt
pkeugine: Wow, very detailed support on your thought 👍 

I can see that it felt natural for you, and you've convinced me on this.
However like you said, since it doesn't have any Lotto-specific behavior,
It could be better if it gets maintained in another utility class or change the naming of `LottoUtils`.
Since we're trying to extract it for later-reuse, we don't want it to be tied to any `Lotto` naming or packaging.

As for the second review, I actually aimed for something like `recursive method call` or `using boolean flag` instead of `while (true)`, but your way could also work!
I would recommend trying to avoid `while (true)` but the way of doing it is completely up to you.
Try all different things and experience them 👍

src/main/kotlin/lotto/domain/WinningCombination.kt
pkeugine: Great.
As for the tip, for now I would recommend trying them out as we go through the missions.
I think in this point, searching and applying the knowledge in the project by him/herself
is the best and quickest way to understand.

Try to apply the knowledge in the project as much as possible.
I'll tell you if something goes wrong, or I feel that it's too far.

src/main/kotlin/lotto/domain/Lotto.kt
pkeugine: I think it would be better for you to think about it and tell me what you thought of.
Then, we'll discuss about that :)

src/main/kotlin/lotto/exceptions/LottoException.kt
pkeugine: I wouldn't say this is over engineering, and there's no one way to manage exceptions.
It would be better for you to try it out yourself, and experience it :)
Then, maybe we'll discuss about it in the next step.

pkeugine (approved):
Hello Sebas,
Nice work, and your responses to my review were great 💯 

Since we have to keep on going with our mission,
I'm gonna merge this PR with responses to some of your questions.
I left some questions not completely answered, because I thought
experiencing it yourself will be much better than me just sharing my thoughts directly.

I'll see you soon, and as always,
Happy hacking 👍 👍 👍

PR #9, opened 2025-07-03
[Lotto - Step 1] Submit arekiu's mission

arekiu:
# Implementation of Lotto simulator

## Features
- Create Lotto Machine
- Receive user input (budget amount) to create tickets.
- Generate tickets with random numbers and display them
- Ask user to input winning and bonus numbers to compare with actual tickets
- Calculate and show winning stadistics and return rate

## Personal decisions and things to review

- Because is not specified in the subject, we decided that the money amount given by user should be a multiple of 1000 (ticket price)
- There´s a warning alert for using the method values() instead of entries() but we didn´t change it because it was like that in the subject
- I didn´t separate the files in Folders or Packages because it´s a small project.
- I would like to know if the responsabilities of Model, View, Controller are well distributed
- I tried to use constants for all the numbers and strings. But it was not implemented in ResultView because of time
- I´m not sure if it´s the right thing, giving almost all the functionalities to LottoMachine class.
- I´m not sure if I should store values in classes attributes instead of creating the variables in main()


README.md
```diff
@@ -1 +1,30 @@
...
+# lotto Features:
+
+- Validate User input Amount:
+- [x] ask user input
+- [x] Ticket purchase should be multiple of 1000
+
+- Generate Tickets:
+- [x] Calculate number of tickets
+- [x] Generate 6 ticket numbers
+  - [ ] Validate non duplicate numbers
```
Wordbe: Not yet implemented! 🙂

gradlew.bat
```diff
@@ -1,92 +1,92 @@
...
-set EXIT_CODE=%ERRORLEVEL%
-if %EXIT_CODE% equ 0 set EXIT_CODE=1
-if not ""=="%GRADLE_EXIT_CONSOLE%" exit %EXIT_CODE%
-exit /b %EXIT_CODE%
-
-:mainEnd
-if "%OS%"=="Windows_NT" endlocal
-
-:omega
+@rem
```
Wordbe, line 1: It doesn't seem like you wrote this gradle.bat file yourself — why was it modified? 🤓

Shall we find out when this file is actually needed?

src/main/kotlin/lotto/InputView.kt
```diff
@@ -0,0 +1,52 @@
...
+                validator.validateBonusNumber(bonusInput, winningNumbers)
+                return bonusInput.toInt()
+            } catch (e: IllegalArgumentException) {
+                println(e.message)
+            }
+        }
+    }
+
+    companion object {
+        const val ASK_AMOUNT = "Please enter the purchase amount."
```
Wordbe: Good job using constant. 🙂

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -0,0 +1,27 @@
...
+        require(lottoNumbers.count() == 6)
+        require(isInRange(lottoNumbers))
+        require(isDuplicated(lottoNumbers))
+    }
+
+    fun getNumbers(): List<Int> {
+        return lottoNumbers
+    }
+
+    private fun isInRange(lottoNumbers: List<Int>): Boolean {
```
Wordbe: How about improving readability with an early return? 🙂

And taking it a step further, maybe you could simplify it using Kotlin’s built-in function like all()?

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,94 @@
+package lotto
+
+class LottoMachine() {
+    private fun calculateTickets(userAmount: Int): Int {
+        val numberOfTickets = userAmount / LOTTO_PRICE
+        return numberOfTickets
```
Wordbe: How about considering inline variables (single-line assignments)? 🤓

Since the method name already clearly describes the return value, that could work well!

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    private fun generateNumbers(): List<Int> {
+        val randomNumbers = (MIN_VAL..MAX_VAL).shuffled().take(6).sorted()
```
Wordbe: Same goes for this one too 🙂

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,94 @@
...
+        return randomNumbers
+    }
+
+    fun createTickets(userAmount: Int): List<Lotto> {
+        val lottos = mutableListOf<Lotto>()
```
Wordbe: Kotlin offers many built-in functions that support functional programming 🙂

How about actively using them to refactor your code in a more functional style?

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,94 @@
...
+        var count = 0
+        while (count < amountOfTickets) {
+            val lotto = Lotto(generateNumbers())
+            lottos.add(lotto)
+            count++
+        }
+        return lottos
+    }
+
+    private fun compareTicketToWinningNumbers(
```
Wordbe: same here!
and all the other methods in here. I suggest you to change these the functional style. 🙂

src/main/kotlin/lotto/Rank.kt
```diff
@@ -0,0 +1,22 @@
...
+    FIRST(6, 2_000_000_000, false),
+    SECOND(5, 30_000_000, true),
+    THIRD(5, 1_500_000, false),
+    FOURTH(4, 50_000, false),
+    FIFTH(3, 5_000, false),
+    MISS(0, 0, false),
+    ;
+
+    companion object {
+        fun valueOf(
```
Wordbe: How about using a different method name instead of valueOf? 🙂

Since it's already a built-in name for enums, it could cause confusion.

src/test/kotlin/LottoMachineTest.kt
```diff
@@ -0,0 +1,35 @@
+package lotto
+
+import org.junit.jupiter.api.Test
+import kotlin.test.DefaultAsserter.assertEquals
```
Wordbe: You're using both JUnit Jupiter and kotlin.test together.
JUnit Jupiter also provides functions like assertEquals 🙂
Let’s think about what might be better for teamwork in this case.

src/test/kotlin/LottoMachineTest.kt
```diff
@@ -0,0 +1,35 @@
...
+
+class LottoMachineTest {
```
Wordbe: When testing LottoMachine, is it possible to configure specific Lotto numbers for the test? 🙂
It’s ideal for tests to be deterministic, where fixed inputs produce expected outputs.

Since the randomness is handled internally, this might be tricky — so how could we refactor the structure to make it easier?

Wordbe (changes requested):
Hello, nice to meet you! 😊
Thank you for your thoughtful code and questions.
I’ve added comments — please take a look and apply the feedback when you can!

---

Q: There´s a warning alert for using the method values() instead of entries() but we didn´t change it because it was like that in the subject

There’s no need for that — just go with whatever looks better 🙂


Q: I didn´t separate the files in Folders or Packages because it´s a small project.

Yes, that’s fine! It’s your choice, but organizing simple packages could be a good idea 🙂

Q: I would like to know if the responsabilities of Model, View, Controller are well distributed

Yes, it looks pretty good! 

Q: I´m not sure if it´s the right thing, giving almost all the functionalities to LottoMachine class.

Exactly, putting everything inside LottoMachine isn’t ideal.
So, how about grouping related functions together and creating new objects with clear responsibilities to separate those concerns? 🙂



Q: I´m not sure if I should store values in classes attributes instead of creating the variables in main()
Objects can hold state and manage related values more easily, which increases cohesion 🙂
It would be great to consider these advantages when making your decision!

src/main/kotlin/lotto/Lotto.kt
arekiu: Thanks for the suggestion. I don´t know yet the different built-in functions that Kotlin offers, and now the change is amazing!

src/main/kotlin/lotto/LottoMachine.kt
arekiu: Now it´s converted to one-liner function and it looks very elegant :D

src/main/kotlin/lotto/LottoMachine.kt
arekiu: Now I learned about the List constructor. I didn´t convert it to one-liner function and I left a variable for better readability

src/main/kotlin/lotto/LottoMachine.kt
arekiu: Now the functions looks much better, I will see what else can I improve in other files

src/main/kotlin/lotto/Rank.kt
arekiu: True! It was like that in the example in the subject, but now I changed it to something more clear "getRank"

src/test/kotlin/LottoMachineTest.kt
arekiu: Tests replaced with AssertEquals from Junit

src/test/kotlin/LottoMachineTest.kt
arekiu: Done! Now there´s a function that takes numbers to create the tickets, instead of generating them inside

README.md
arekiu: I realized that is difficult to test those generated numbers because they are random. But to validate that they are duplicated and unique, I can trust the functionality of shuffled().take()
On the other hand, the Lotto ticket won´t be created if the numbers are not fulfilling the conditions.
I´m just going to delete those points from the feature list

gradlew.bat
arekiu: The project was done in the laptop (Mac) of my peer, I just copied the src folder and the Readme to add them into my project, and I never touched that file.

The file is part of Gradle wrapper.  Is needed for consistency to build the project in different systems.
A peer had the same situation and it seems it´s because .gitattributes interprete there´s a change when there´s a different end-of-line setting.

arekiu:
Thank you so much for the detailed review! And nice to meet you too!

Here is the refactored version of my project. I created packages to organize the files and more classes to distribute the functionalities that were part of the LottoMachine.

README.md
Wordbe: The current LottoManager is responsible for generating random lotto numbers. If we extract just that part to the outermost layer, it would become testable. If you have time, would you like to give it a try? 😊

gradlew.bat
Wordbe: okay that's good 🙂

src/main/kotlin/lotto/Lotto.kt
Wordbe: Great job 👏

src/main/kotlin/lotto/LottoMachine.kt
Wordbe: Yes Kotlin users love this style!

src/main/kotlin/lotto/LottoMachine.kt
Wordbe: Okay 👌 good!

src/main/kotlin/lotto/Rank.kt
Wordbe: Looks good 🙂

Wordbe (changes requested):
Thank you so much for your thoughtful feedback and for reflecting on the points I mentioned earlier — I really appreciate it!

There’s just one thing I’d like to suggest modifying: the way member variables are declared.
In Kotlin, it’s common practice to declare member variables in the primary constructor.
Looking at classes like LottoMachine and Lotto, I thought it would be great to make more active use of the primary constructor.

By using the primary constructor:

The code becomes more concise, Getters and setters are automatically generated, You can perform validations during initialization to create safer objects, And constructor-based dependency injection becomes much more natural. 

Would you consider trying that approach?
Also, instead of manually writing getters and setters, please try taking advantage of the approach I mentioned above.

Thanks again! 😊

README.md
arekiu: Now I created a separated class "NumberGenerator" to take care of it

arekiu:
Thank you again for your nice review.

While trying to modify LottoMachine to set values in the primary constructor, I realized that the structure was not so good.
So I decided to create new classes to process the input and then initialize LottoMachine with it.
In my new software design, there´s a LottoSeller that calculate the tickets, and there´s a data class PrizeNumber that stores winning and bonus numbers. 
When this information is given to LottoMachine, this class will be in charge of calculate the results (with the help of LottoManager).

I also created dedicated classes to generate the random numbers, and for Ticket creation.

Thank you for your time!

src/main/kotlin/lotto/controller/LottoController.kt
```diff
@@ -0,0 +1,39 @@
...
+import lotto.model.TicketsSeller
+import lotto.view.InputView
+import lotto.view.ResultView
+
+class LottoController {
+    private val inputView = InputView()
+    private val resultView = ResultView()
+    private val ticketSeller = TicketsSeller()
+
+    private var userAmount = 0
```
Wordbe, line 14: The `LottoController` currently holds a state through the `userAmount` variable.

I recommend thinking about why maintaining state here might not be ideal, and try refactoring the structure to remove it! 😊

src/main/kotlin/lotto/view/ResultView.kt
```diff
@@ -0,0 +1,39 @@
...
+class ResultView() {
+    fun printLottos(lottos: List<Lotto>) {
+        println("You have purchased ${lottos.count()} tickets.")
+        lottos.forEach { lotto ->
+            val lottoNumbers = lotto.lottoNumbers
+            println(lottoNumbers.joinToString(",", "[", "]"))
+        }
+    }
+
+    fun printResult(results: MutableMap<Rank, Int>) {
```
Wordbe, line 15: How about combining printResult and printReturnRate into a single method?
It would make the result output more cohesive and easier to manage. 😊

src/main/kotlin/lotto/view/ResultView.kt
```diff
@@ -0,0 +1,39 @@
...
+        println()
```
Wordbe, line 16: How about combining the println statements starting from here into a single block using `"""`?
It would make the output cleaner and easier to read. 😊

src/main/kotlin/lotto/model/Validator.kt
```diff
@@ -0,0 +1,51 @@
...
+            throw IllegalArgumentException("Numbers should be unique")
+        }
+        userInput.forEach { item ->
+            if (item.toInt() !in 1..45) {
+                throw IllegalArgumentException("The winning numbers should be between 1 and 45")
+            }
+        }
+    }
+
+    fun validateBonusNumber(
```
Wordbe, line 32: It might be a good idea to manage exceptions more easily using require() or check(). 😊

(Also other validations in the Validator)

src/main/kotlin/lotto/model/Lotto.kt
```diff
@@ -0,0 +1,38 @@
...
+    private fun numbersAreNotDuplicated(lottoNumbers: List<Int>): Boolean {
+        return lottoNumbers.count() == lottoNumbers.toSet().count()
+    }
+
+    fun compareTicket(winningNumbers: List<Int>, bonusNumber: Int) {
+        compareTicketToWinningNumbers(winningNumbers)
+        if (numberOfHits == 5) compareTicketToBonusNumber(bonusNumber)
+    }
+
+    private fun compareTicketToWinningNumbers(
```
Wordbe, line 27: It would be better if this method returned the result directly instead of holding numberOfHits as a state! 😊

src/main/kotlin/lotto/model/Lotto.kt
```diff
@@ -0,0 +1,38 @@
...
+        winningNumbers: List<Int>,
+    ) {
+        numberOfHits = winningNumbers.count { it in lottoNumbers }
+    }
+
+    private fun compareTicketToBonusNumber(
```
Wordbe, line 33: The same goes for `bonusNumber` — it’d be better to return it directly instead of keeping it as state! 😊

Wordbe (approved):
Thank you for thoughtfully reflecting the review feedback!

I’ve added a few more suggestions that could be improved.
I’ll go ahead and merge this PR now to keep things moving for the next mission, so it would be great if you could naturally incorporate those points at the beginning of the next one.

Thanks again for all your hard work on the Lotto mission - great job! 😊

src/main/kotlin/lotto/controller/LottoController.kt
arekiu: Ok, I understood that I can be problematic. Now I removed it and transformed it to a local variable.

src/main/kotlin/lotto/view/ResultView.kt
arekiu: Done! I created another method, so I call printTicketResults and printReturnRate inside printResult

PR #10, opened 2025-07-03
[Lotto - Step 1] Submit knopgm's mission

knopgm:
### **What**
Implements a Lotto Machine:

- Users buy tickets based on budget input.
- System generates random tickets (6 numbers each).
- Users input winning numbers + bonus number.
- Tickets are checked against results and prizes are calculated.
- Final profit rate is displayed.

### **Why**
As a task of the step1 in Hero Tech Course, to better understand TDD, code structure, learn and practice SOLID principles on top of practicing to listening and argue with peers while pair programing or inside teams.

### **How**

- InputView and ResultView manage user interaction.
- LottoMachine manage all logic of the lotto game.
       - Calculate and generate tickets
       - Compare bonus and winning numbers with ticket numbers
       - Calculate total prizes and profit rate


README.md
```diff
@@ -1 +1,30 @@
...
+- Validate User input Amount:
+- [x] ask user input
+- [x] Ticket purchase should be multiple of 1000
+
+- Generate Tickets:
+- [x] Calculate number of tickets
+- [x] Generate 6 ticket numbers
+  - [ ] Validate non duplicate numbers
+  - [ ] Numbers should be between 1 and 45
+- [x] Create tickets with generated numbers
```
pkeugine: Nice job organizing the feature list in README 👍 
Make sure to check the works that you've done.
I recommend making it a habit to go through the list before making the commit to make sure the check goes along with the feature. 🙂

The commits also look nice, but to improve it,
I recommend git's `amend` feature and `rebase` feature.
By using them, you'll be able to rename the `WIP` commits or put together multiple commits in one whole commit. It'll allow us to stack those commits with more completeness.

There's also git features you can use in intelliJ.
It'll allow you to `amend` and `rebase` commits more easily.
It's worth checking it out :)

build.gradle.kts
```diff
@@ -18,6 +18,7 @@ dependencies {
     testImplementation("org.junit.jupiter", "junit-jupiter", "5.10.2")
     testImplementation("org.assertj", "assertj-core", "3.25.3")
     testImplementation("io.kotest", "kotest-runner-junit5", "5.8.0")
+    testImplementation(kotlin("test"))
```
pkeugine: I see you’ve added a new dependency to the project 😮
Although it’s used in the tests, I think we can achieve the same result using what’s already provided without introducing a new dependency.
**Hint**: [assertJ](https://assertj.github.io/doc/)

Even though it's written in Java, it shouldn’t be too difficult to use,
especially if you check out the documentation I linked above.
It’s actually one of the best resources to start reading when learning programming.

src/main/kotlin/lotto/Application.kt
```diff
@@ -0,0 +1,22 @@
...
+
+    // Compare winning tickets from given winning numbers
+    val winningNumbers = inputView.takeWinningNumbers()
+    val bonusNumber = inputView.takeBonusNumber(winningNumbers)
+    val results = lottoMachine.compareTickets(tickets, winningNumbers, bonusNumber)
+    resultView.printResult(results)
+
+    // Calculate profit rate
+    val rate = lottoMachine.calculateReturnRate(results, userAmount)
+    resultView.printReturnRate(rate)
```
pkeugine: You've put in a lot of work in readability when writing application logic. 💯 
It's so great that it feels unnecessary to add these comments in between.
I believe great code explains itself, and you've done that.
So why don't we remove these comments and let the code express themselves? 😄 

Some might ask, `then when do we actually add comments?`
I want you to take some time to think about this and tell me your thoughts on it.
This question doesn't have a one right answer so take it lightly. :)

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -0,0 +1,27 @@
+package lotto
+
+class Lotto(val lottoNumbers: List<Int>) {
+    init {
+        require(lottoNumbers.count() == 6)
+        require(isInRange(lottoNumbers))
+        require(isDuplicated(lottoNumbers))
```
pkeugine: Like the previous review, naming methods and variables is important for logic to express itself.
It helps collaborating with others or maintaining a project for a long time.

When reading this part, for example, the first two requirements feel natural to me.
However, the third requirement `require(isDuplicated...` makes me think `So the Lotto needs duplicated numbers?`.
I go into the implementation and see that it's the opposite.
Why don't we change the method name so that it reflects on what it really validates?

src/main/kotlin/lotto/ResultView.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    fun printReturnRate(rate: Double) {
+        val formattedRate = String.format("%.1f", rate).replace(',', '.')
+        println("Total return rate is $formattedRate")
+    }
+
+    companion object Messages {
+        const val PRESENT_WINNING = "Winning Statistics"
+        const val SEPARATOR = "------------------"
+    }
```
pkeugine: I see you've substituted some messages with constants while keeping some as it is in the logic.
Makes me wonder: `Which ones do you substitute, and which ones do you keep it as is?`
I'm guessing it's about using the `results.getValue`.

Also, I personally try to avoid keeping console messages or error messages as constants. I'm not saying it's wrong since there are developers who prefer this way. Let's have a moment to think about why I try to avoid it, and why might one prefer it.

src/main/kotlin/lotto/Validator.kt
```diff
@@ -0,0 +1,51 @@
...
+        }
+        if (userInput.count() != userInput.toSet().count()) {
+            throw IllegalArgumentException("Numbers should be unique")
+        }
+        userInput.forEach { item ->
+            if (item.toInt() !in 1..45) {
+                throw IllegalArgumentException("The winning numbers should be between 1 and 45")
+            }
+        }
+    }
```
pkeugine: Excellent validating for all different kinds of domains. 👍 
Instead of keeping them in `Validator`, why don't we create classes and give them the responsibility to validate it themselves?

For example, we can make a `LottoNumber` class and let it validate if the number is in between 1 and 45.
By doing so, we'll be able to use it for bonus number as well!

src/main/kotlin/lotto/InputView.kt
```diff
@@ -0,0 +1,52 @@
...
+                println(ASK_WINNING_NUMBERS)
+                val input = readln().split(",")
+                input.forEach { item -> item.trim() }
+                validator.validateWinningNumbers(input)
+                return input
+            } catch (e: IllegalArgumentException) {
+                println(e.message)
+            }
+        }
+    }
```
pkeugine: It's pretty common for users to accidentally (or even intentionally...!) mess around with inputs when using your application like this:

Screenshot 2025-07-04 at 11 50 11 PM

Some inputs are actually unacceptable, but I think above situation could be covered and be treated as acceptable input. Why don't we try it?

This gives me an idea that we can also create tests about it. 😄

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,94 @@
...
+        winningNumbers: List<String>,
+    ): Int {
+        var countMatches = 0
+        winningNumbers.forEach { number ->
+            if (number.toInt() in lotto.getNumbers()) {
+                countMatches++
+            }
+        }
+        return countMatches
+    }
```
pkeugine: Why don't we order the methods in classes so that the order actually contain meaning?
For example, one might put public methods before private ones.
Some might put private methods below the method that uses it.

I would like you to set up your own convention about this and apply it to the project. It'll be fun to talk about this with your pair or even fellow crews and see how they order the methods 🙂

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,94 @@
...
+    private fun compareTicketToBonusNumber(
+        lotto: Lotto,
+        bonusNumber: Int,
+    ): Boolean {
+        return bonusNumber in lotto.getNumbers()
+    }
+
+    private fun createMap(): MutableMap<Rank, Int> {
+        val prizeCounter = mutableMapOf<Rank, Int>()
+        for (rank in Rank.values()) {
```
pkeugine:
```suggestion
        for (rank in Rank.entries) {
```

We can try using this instead.
I would like you to find the difference between these two and tell me about it :)

src/main/kotlin/lotto/Rank.kt
```diff
@@ -0,0 +1,22 @@
...
+    ;
+
+    companion object {
+        fun valueOf(
+            countOfMatch: Int,
+            matchBonus: Boolean,
+        ): Rank {
+            return values().find { it ->
+                (it.countOfMatch == countOfMatch && it.hasBonus == matchBonus)
+            } ?: MISS
```
pkeugine:
```suggestion
            return entries.find {
                it.countOfMatch == countOfMatch && it.hasBonus == matchBonus
            } ?: MISS
```

Just like the above review, we can use `entries` and remove the arrow for lambda expression here.

src/main/kotlin/lotto/ResultView.kt
pkeugine: Additional note: there's also a [multi-line String](https://kotlinlang.org/docs/java-to-kotlin-idioms-strings.html#use-multiline-strings) option here :)

src/test/kotlin/LottoMachineTest.kt
```diff
@@ -0,0 +1,35 @@
...
+        assertEquals("Amount of tickets is different then expected", lottoMachine.createTickets(4000).count(), 4)
+    }
+
+    @Test
+    fun `correct amount of rate is calculated`() {
+        val lottoMachine = LottoMachine()
+        val results = mutableMapOf<Rank, Int>()
+        results[Rank.FIFTH] = 1
+        val amount = 3000
+        assertEquals("Rate result is different then expected", lottoMachine.calculateReturnRate(results, amount), 1.6666666666666667)
```
pkeugine:
```suggestion
        assertEquals(
            actual = lottoMachine.calculateReturnRate(results, amount),
            expected = 1.66666666666667,
            message = "Rate result is different than expected"
        )
```

Although it might look like they bring the same results,
putting values into their respected place is important for logic to work as expected.

One way to achieve this is by using [named parameters](https://kotlinlang.org/docs/functions.html#named-arguments) just like the example above.
Before it, `1.6666666666666667` was in the `actual` spot and `lottoMachine.calculate...` was in the `actual`. Also, notice that I put the message at the very bottom since it's the least important part of this assertion.

Let's try refactoring the test logic :)

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,94 @@
...
+
+        lottos.forEach { ticket ->
+            var hasBonus = false
+            val matches = compareTicketToWinningNumbers(ticket, winningNumber)
+            if (matches == 5) {
+                hasBonus = compareTicketToBonusNumber(ticket, bonusNumber)
+            }
+            val chosen = Rank.valueOf(matches, hasBonus)
+            prizeCounter[chosen] = prizeCounter.getValue(chosen) + 1
+        }
```
pkeugine: I think we can move this logic into `Lotto` or `Rank` and let them be responsible for calculating the rank.

Easy way to think about it:
Try to use `get...()` kind of getters only for output view purposes. If you need to take out values from particular object, chances are, that object should be the one calculating it.

There are times where you HAVE to use getters or there must be a massive refactoring throughout the whole project. In that case, it's fine to use getters I guess. But if that's not the case, let's try keeping above practice.

src/test/kotlin/RankTest.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.junit.jupiter.api.Test
+import kotlin.test.DefaultAsserter.assertEquals
+
+class RankTest {
+    @Test
+    fun `should assign the corresponding rank when it has bonus`() {
+        val matchingNumber = 5
+        val hasBonus = true
+
+        assertEquals("Should match the right rank", Rank.valueOf(matchingNumber, hasBonus).toString(), "SECOND")
```
pkeugine:
```suggestion
        assertEquals("Should match the right rank", Rank.valueOf(matchingNumber, hasBonus), Rank.SECOND)
```
Since we're using `assertEquals`, we can directly compare enum values to check if we're getting the expected result.

Why does this work?
What's special about `enum` classes in Kotlin (and Java), and what does it actually mean when two enum values are considered `"equal"`?

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    fun getNumbers(): List<Int> {
+        return lottoNumbers
+    }
+
+    private fun isInRange(lottoNumbers: List<Int>): Boolean {
+        var isInRange = true
+        lottoNumbers.forEach { number ->
+            if (number !in 1..45) {
```
pkeugine: The numbers `1` and `45` hold special meaning in this application : They’re the boundary values of a lotto ticket. We often call these 🔮 [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) 🔮. Since they’re important, it’s a good idea to extract them as constants with meaningful names, like:
```kotlin
const val MINIMUM_LOTTO_NUMBER = 1
const val MAXIMUM_LOTTO_NUMBER = 45
```

Additionally, we could consider creating a `LottoNumber` class to encapsulate validation logic for this range. Let’s give it a try! 😄 

---

I found `MAX_VAL` and `MIN_VAL` in `LottoMachine`. Well done 👍 
However, it'll be nice if we can use that constant across the whole project,
so 1 and 45, when used as lotto number, gets substituted with that constant.

To make it more easy, putting all the logics related to these number in `LottoNumber` might help :)

src/main/kotlin/lotto/InputView.kt
```diff
@@ -0,0 +1,52 @@
...
+            try {
+                println(ASK_BONUS_NUMBER)
+                val bonusInput = readln()
+                validator.validateBonusNumber(bonusInput, winningNumbers)
+                return bonusInput.toInt()
+            } catch (e: IllegalArgumentException) {
+                println(e.message)
+            }
+        }
+    }
```
pkeugine: `while (true)` can be a dangerous move. It often leads to bugs no one intended.  
It may look unharmful for now, but trust me, I've spent a lot of Friday nights with this guy 🥲 

Well, yes, I totally get that using a `while (true)` loop to keep asking for valid user input is a classic and valid use case, and many developers advocate for it.

Still, there are other options available to us, so it’s worth thinking about what those are, and whether they might be a better fit than a `while (true)` loop 🤓

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,94 @@
...
+        userAmount: Int,
+    ): Double {
+        val totalPrize = calculateTotalPrize(results)
+        return (totalPrize.toDouble() / userAmount.toDouble())
+    }
+
+    companion object {
+        const val LOTTO_PRICE = 1000
+        const val MIN_VAL = 1
+        const val MAX_VAL = 45
```
pkeugine: `MIN` and `MAX` seems fine since they get used in daily basis, but let's try not to shorten names. Shortened names often come down to personal preference, which can lead to inconsistent naming across the codebase.

src/test/kotlin/RankTest.kt
```diff
@@ -0,0 +1,30 @@
...
+        assertEquals("Should match the right rank", Rank.valueOf(matchingNumber, hasBonus).toString(), "THIRD")
+    }
+
+    @Test
+    fun `should assign MISS when is not matching less than 3 numbers`() {
+        val matchingNumber = 2
+        val hasBonus = false
+
+        assertEquals("Should match the right rank", Rank.valueOf(matchingNumber, hasBonus).toString(), "MISS")
+    }
```
pkeugine: I'm curious about your experience with TDD.
In which part was it helpful, and where did it feel hard?

pkeugine (changes requested):
Hello Gabi, and congrats on your first PR! 🍾 
The code was really easy to read, and reviewing it was a breeze. 💯 

I added some reviews with my personal thoughts along with suggestions and questions that could help you learn more as we go through this mission.

I’ll mark this PR as "changes requested" for now.
Once you’ve gone through the reviews and made updates, feel free to send another review request and we’ll go over the changes together.

If you have any questions, feel free to leave a comment on this PR or DM me anytime.
Happy hacking! ⌨️ ⌨️ ⌨️

README.md
knopgm: Hi pKeugine,
Thank you for such a nice review! I'm just starting to do all your recommendations as better as I can.
One fact that is worrying me the most is the short deadlines we need to reach. These is something I took in consideration all the way through and is showing up in some requests you made me. 
I checked the readme tasks but I'm too afraid to touch GitHub amend and rebase now. 
Is it a suggestion for future commits or should I amend those for the completion of this task?

src/main/kotlin/lotto/Application.kt
knopgm: Thank you for ensuring my code is readable! It's a relief and a pleasure to know that I'm on the right path! :)
I remove the unnecessary comments.
Thinking about your question, I remember reading that comments are great to clarify complex connections, non obvious code or explain why we are using something, but since I'm a beginner it isn't obvious to me (yet..) when the code is clear enough. Maybe is because I didn't code in a big project yet.
Nevertheless, the comments you point were though to be a good "Title" for the application divisions of logic.
I'll get use not to make titles for my codes. 😄

src/main/kotlin/lotto/ResultView.kt
knopgm: Thank you for bringing these questions up. In my research it says that avoid keeping console messages as constants in small projects or non-reusable messages can improve readability and avoid over complexity, keeping it easier to maintain. Although it should be a good practice if messages are being used in another parts of the project or in large projects, which can help maintenance. :)

src/main/kotlin/lotto/Validator.kt
knopgm: I thought about doing that (a validator for each responsibility), but since we had a tight deadline,  I leaved it to be a refactor approach.

src/main/kotlin/lotto/LottoMachine.kt
knopgm: Great topic! I had never thought about which order the function types should follow. Now it makes sense that the public function, as the one that is connecting the external data to the logic, comes before the private ones. Thank you for bringing this to my attention! xD

src/main/kotlin/lotto/InputView.kt
knopgm: I also had bad situations with while loops before, but I can't think of another good solution here. 
Is a do while in someway a better solution? For me it looks the same 🥲

src/test/kotlin/RankTest.kt
knopgm: We had only a clear benefit from TDD when we create the test for calculating the correct amount of rate. Since the correct answer was a float number, we saw the failed test because we didn't used the correct amount of numbers after the dot (1.66666666... hard to guess how many 6 it has) so we copy the correct number directly from console error message.
But I guess it is normal for TDD beginners to think it is hard to test before coding the functions. We still have difficulty in visualize all features and connections, and we are also very new to test libraries.
Nevertheless, I understand it's power of ensuring the safety of code and will keep practicing until it gets a habit. 😅

src/main/kotlin/lotto/LottoMachine.kt
knopgm: I moved the comparison of winning numbers and bonusNumbers to Lotto class, since it alredy has the lotto tickets to compare with this numbers.
About the get( ), I can't think about other solution to build the prizes list to print (that comes from inside a map where I save all matched tickets. But I'll keep in mind that get() is not a good thing to use.
I'm also open to refactor it again! Let me know what can I do instead of this get (and consequently the way I saved the matched prizes). Thank you so much! 🙏

src/main/kotlin/lotto/Validator.kt
knopgm: I rename this validator as InoutValidator, since it is resposible for validating all use inputs. There are other classes that are validate themselves like Lotto and LottoNumber classes that are starting with a init{ require(...) }, maybe I'm confused about when to use one or the other... but is working well for my deadline of this task 😅

src/test/kotlin/RankTest.kt
knopgm: In my research enum classes are objects, so comparing it as objects (their actually location on the memory) is better (memory-safe) than comparing two strings. Enum classes are special because they hold a fixed set of constant values which are all a singleton object. In Kotlin we can define properties and functions for each enum value, and also override  or create custom factory methods. Thanks for made me research those concepts again! 😄

knopgm:
Hi @pkeugine,
I'm super happy with your reviews! All very detailed and with a lot food for thought and for learn :)
I try to address all of the reviews. Some of them I couldn't understand and I'll bring to class tomorrow. 
Feel free to require more improvements when you think it may fit.
Thank you so much! 🙏

README.md
pkeugine: Oh yes, as much as learning, **deadlines** are also very important.
And I also agree that git skills that I've mentioned above need some time to practice
as they are not something one obtains in a day just by reading the documents.

So it's fine to pass this for now, and like you said, keep in mind when committing future tasks. 👍

src/main/kotlin/lotto/Application.kt
pkeugine: Nice 👍 
And yes, comments are great when clarifying complex connections, and they also give you examples and notifications of how to utilize that class (or method).

One way to experience comments is to check the implementations of libraries or programming language api.
Java, Kotlin, AssertJ, ...etc they all contain comments that are actually very helpful. (They contain similar information to their official documentation).

src/main/kotlin/lotto/LottoMachine.kt
pkeugine: We share the same idea :)
me : `Try to use get...() kind of getters only for output view purposes.`
Gabi : `I can't think about other solution to build the prizes list to print`

In my early days of programming, I actually tried not to use `get()` even for output purposes,
but... there wasn't any other way to do it 😂

src/main/kotlin/lotto/InputView.kt
pkeugine: I think we can try out recursive method calls or using boolean local variable as a flag, rather than keeping it always true.
However, if you're running close to deadline, I recommend thinking about this comment later on.

src/test/kotlin/RankTest.kt
pkeugine: Aha, I see that the upside was very practical :)
And yes, it's really hard to maintain TDD even for a programming veteran especially if we're working on a new domain.
Since you weren't used to the testing library, it must've been much harder. Still, you did a great job 💯 

Next time, it'll be nice for you to focus on the variety of tests,
such as more fixtures, more cases, and even organizing those increased amount of tests in your own convention.
By using testing libraries, it'll be pretty fun. Also, it's not always about the tools : by just making test name conventions, the whole team will be able to organize and utilize tests in a much easier way.

gradlew.bat
```diff
@@ -1,92 +1,92 @@
...
+@rem ##########################################################################
+
+@rem Set local scope for the variables with windows NT shell
+if "%OS%"=="Windows_NT" setlocal
+
+set DIRNAME=%~dp0
+if "%DIRNAME%"=="" set DIRNAME=.
+@rem This is normally unused
+set APP_BASE_NAME=%~n0
+set APP_HOME=%DIRNAME%
```
pkeugine, line 31: I skipped this review at first, but I think it would be nice for you to think about it now.
This `gradle.bat` file is not something you've written or changed, but it's in this PR as 'changed file'.
Why do you think this happened? Also, what IS `gradlew.bat` anyway? 🤔

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -0,0 +1,66 @@
...
+
+class Lotto(val lottoNumbers: List<LottoNumber>) {
+    init {
+        require(lottoNumbers.count() == 6)
+        require(isNotDuplicated(lottoNumbers))
+    }
+
+    fun getLottoNumbersAsInt(): List<Int> {
+        return lottoNumbers.map { it.value }
+    }
```
pkeugine, line 13: If we start to utilize Kotlin more and more, we'll see there's so many simple-looking-ways of achieving the same goal :

```kotlin
val lottoNumbersAsInt: List<Int>
    get() = lottoNumbers.map { it.value }
```

src/test/kotlin/LottoNumberTest.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+
+class LottoNumberTest {
+    @Test
+    fun `throw error exception when number is out of the range between 1 and 45`() {
+        for (i in 1..45) {
+            LottoNumber(i)
+        }
```
pkeugine, line 11: This for-loop in a test is not asserting anything.
If this object gets created without any exceptions, let's assert that with our test library 👍

src/test/kotlin/LottoMachineTest.kt
```diff
@@ -0,0 +1,53 @@
...
+    fun `correct amount of rate is calculated`() {
+        val lottoMachine = LottoMachine()
+        val results = mutableMapOf<Rank, Int>()
+        results[Rank.FIFTH] = 1
+        val amount = 3000
+        assertEquals(
+            lottoMachine.calculateReturnRate(results, amount),
+            1.6666666666666667,
+            "Rate result is different than expected",
+        )
```
pkeugine, line 51: I know that there's a deadline and things might be in a rush, [but don't forget to apply reviews](https://github.com/herotechcourse/kotlin-lotto/pull/10#discussion_r2185627847).
It's also important not to mark comments as `resolved` even though it hasn't been.
It confuses the readers and results in longer review time.
If you think it's better to keep it this way or have a different thought,
teammates and reviewers are always welcome to hear it.

pkeugine (approved):
Hello Gabi,
Well done with your first step of the mission.
I'll merge this PR right away so we can continue on with our mission.

I've left some comments so it'll be nice for you to check them out as you work on your next programming.

If you have any questions, feel free to hit me with it in any way.
Nice work, and as always,
Happy hacking 🏃‍♂️🏃‍♂️ 🏃‍♂️

gradlew.bat
knopgm: Good questions!
The answer I found on the internet is that gradlew.bat  is the same as gradlew that is the sript to build the project. The W at the end means Wrapper, and its meant to invoke a declared version of Gradle, downloading it beforehand if necessary. 
The reason why it is changed I don't know. If you can, I would be thrilled to know what happened or at least a clue to where can I found the reason for the change. I try to compare both codes (current and previous - by the log commit) and I couldn't see one changed line in this file. 😅 
Could it be another file changes that affected this one? Because I don't remember to have touched this file.

src/test/kotlin/LottoNumberTest.kt
knopgm: You're right!
Since the LottoTest is already testing the numbers range required to be a lottoTicket, I think this test has no use.
I deleted it.

src/test/kotlin/LottoMachineTest.kt
knopgm: I'm so sorry! Probably I miss understood the review...
I just asked the coaches and now I see what should've done in this review.
Please, let me know if what you meant was that I misplaced the location of what I was expecting from output of the function, and what I actually had (as a mocked result).
Thank you for your patience in going back to alert me of that unresolved issue! 🙏

src/main/kotlin/lotto/InputView.kt
knopgm: I remove the logic from inside my while loop for now (also following coaches instructions towards this issue), but I still think that a while loop to re-input the user during the application run is less stressful and user-friendly then just crashing the application. 😅 
What do you think about it? Should I ignore the UI when I'm building backend?

gradlew.bat
pkeugine: It'll be great if you could search the reason behind it. If it's hard to find, I think there are crews who found out, so I recommend talking with them about this 👍

src/main/kotlin/lotto/InputView.kt
pkeugine: What I intended in this review was to achieve the same kind of UI without `while (true)` loop with recursive method calls or local boolean flag. I do think UI is important even when we (as a backend developer) create it, since UI is part of our job in this case. However, I think other topics are much important so it's fine to leave this out for now :)

src/test/kotlin/LottoMachineTest.kt
pkeugine: The comment could've created some confusion since I said both values were in the `actual` spot, but what I meant was `1.6666666666666667` should be an expected, and `lottoMachine.calculateReturnRate(results, amount)` must be an actual. In the code below, they are in the opposite spot:  
```Kotlin
        assertEquals(
            lottoMachine.calculateReturnRate(results, amount),
            1.6666666666666667,
            "Rate result is different than expected",
        )
```

It'll be easier to understand if you see the method signature of `assertEqual` : 
```Kotlin
    fun assertEquals(message: String?, expected: Any?, actual: Any?): Unit
```

pkeugine: It's fine :)
The comment could’ve created some confusion since I said both values were in the `actual` spot 😅 
What I meant was `1.6666666666666667` should be an `expected`, and `lottoMachine.calculateReturnRate(results, amount)` must be an `actual`. In the code below, they are in the opposite spot:  
```Kotlin
        assertEquals(
            lottoMachine.calculateReturnRate(results, amount),
            1.6666666666666667,
            “Rate result is different than expected”,
        )
```

It’ll be easier to understand if you see the method signature of `assertEqual` : 
```Kotlin
fun <@OnlyInputTypes T> assertEquals(expected: T, actual: T, message: String? = null) {
```

`actual` -> what our logic returns (ex: `lottoMachine.calculateReturnRate(results, amount)`)
`expected` -> the value we set in our test (ex: `1.6666666666666667`)

PR #11, opened 2025-07-03
[Lotto - Step 1] Submit Mariia's mission

mariyakulikova:
Hi!
A pull request add our implementation of console game lotto. It was interesting experience of first time pair programming. We tried to look for a middle ground between implementation complexity and simplicity. 

We would appreciate feedback on following question:

1. What would you change or remove?
2. Are test cases adequate or there is need in additional tests?
3. Where to handle inputed date validation properly?
4. How to choose between object and class properly?


README.md
```diff
@@ -1 +1,34 @@
-# kotlin-lotto
\ No newline at end of file
+# kotlin-lotto
+
+Functional Requirements
```
Wordbe, line 3: That's a neat summary, thank you. 😊

src/main/kotlin/lotto/Calculator.kt
```diff
@@ -0,0 +1,8 @@
+package lotto
+
+object Calculator {
+    fun division(
```
Wordbe: Is there a reason we need to create this feature? 🤔

If it doesn’t include things like exception handling for division by zero, why not try using Kotlin’s built-in functionality?

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -0,0 +1,27 @@
...
+        }
+        require(numbers.toSet().size == NUMBERS_SIZE) {
+            "lotto ticket numbers should be unique"
+        }
+        require(numbers.all { it in 1..45 }) {
+            "lotto ticket each number should be between 1 and 45"
+        }
+    }
+
+    fun matchCount(winningNums: List<Int>): Int {
```
Wordbe, line 18: It’s a simple feature with few elements in the collection, but I’d love for us to try developing with performance in mind at least once!

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -0,0 +1,27 @@
...
+        return numbers.count {
+            it in winningNums
+        }
+    }
+
+    fun containsBonusNum(num: Int) = num in numbers
+
+    open fun getListOfNumbers(): List<Int> = numbers
```
Wordbe: There’s already a getter for numbers—is there a reason we created a new one? 🙂

src/main/kotlin/lotto/LottoGame.kt
```diff
@@ -0,0 +1,17 @@
+package lotto
+
+class LottoGame(val playersTickets: List<Lotto>, val winningNumbers: Lotto, val bonusNumber: Int) {
+    val results: MutableMap<Rank, Int> = mutableMapOf()
```
Wordbe: Instead of holding results and winningAmount as states, could we return them all at once? 🙂

Maintaining state introduces the possibility of changes, adds complexity, and can lead to side effects. I’d love for us to think this through.

src/main/kotlin/lotto/Money.kt
```diff
@@ -0,0 +1,14 @@
+package lotto
+
+class Money(val amount: Int) {
+    init {
```
Wordbe, line 4: Using constant names instead of magic numbers makes management easier and improves readability during reviews. 🙂

src/main/kotlin/lotto/Money.kt
```diff
@@ -0,0 +1,14 @@
...
+        require(amount >= 1000) {
+            "the amount should be minumum 1,000 KRW"
+        }
+        require(amount % 1000 == 0) {
+            "the amount should be divisible by 1000"
+        }
+    }
+
+//    fun getMoneyAmount(): Int = amount
```
Wordbe: Please remove unnecessary comments or unused code. 🧹

src/main/kotlin/lotto/Rank.kt
```diff
@@ -0,0 +1,30 @@
...
+    FOURTH(4, false, 50_000),
+    FIFTH(3, false, 5_000),
+    NONE(0, false, 0), ;
+
+    companion object {
+        fun getRank(
+            matchCount: Int,
+            matchBonus: Boolean,
+        ): Rank {
+            return when {
```
Wordbe: How about using when expression like `when (matchCount)` to improve readabiltiy? 🙂

src/main/kotlin/view/ResultView.kt
```diff
@@ -0,0 +1,39 @@
...
+        println("You have purchased $ticketsAmount tickets")
+    }
+
+    fun viewTickets(tickets: List<Lotto>) {
+        tickets.forEach {
+            println(it.getListOfNumbers().toString())
+        }
+    }
+
+    fun viewWinning(resul: MutableMap<Rank, Int>) {
```
Wordbe: In ResultView, it would be best to avoid complex calculations and simply receive the computed values for display—this could help with clearer separation of responsibilities. 😄
Also, there's a typo in resul!

src/test/kotlin/lotto/LottoGameTest.kt
```diff
@@ -0,0 +1,30 @@
...
+                Lotto(listOf(1, 2, 3, 4, 5, 7)),
+                Lotto(listOf(1, 2, 3, 4, 5, 45)),
+                Lotto(listOf(1, 2, 3, 4, 44, 45)),
+                Lotto(listOf(1, 2, 3, 43, 44, 45)),
+                Lotto(listOf(10, 11, 12, 13, 14, 15)),
+            )
+        val game = LottoGame(tickets, winningNumbers, bonusNumber)
+        game.play()
+        val result = game.results
+        assertEquals(1, result.getValue(Rank.FIRST))
```
Wordbe: Using a Map is convenient, but from a reader’s perspective, it’s hard to tell what the keys and values represent. 

i.e. What is the value? in this line.

If time allows, it would be great to try expressing this with a custom object.

Wordbe (commented):
Hello! Mariia
Thank you for your thoughtful code and questions. 😀
Please take a look at the comments and try reflecting them in your updates!

---

Q1: What would you change or remove?

I’ve noted it in the comments!

Q2: Are test cases adequate or there is need in additional tests?

It would be great if each of the features (methods) had corresponding tests. 
Also, please supplement anything that seems lacking overall.

Q3: Where to handle inputed date validation properly?

The more thorough the validation, the better. 🙂
It can be done at various levels—such as domain objects like Lotto or LottoMachine, or even at the view level like InputView.
Especially when validation happens at the point of domain object creation, it helps ensure data integrity while constructing the object.

Q4: How to choose between object and class properly?

First, it’s good to understand the difference between an object and a class. 🙂
You can mainly distinguish them based on whether they hold state and whether a singleton is needed.

Wordbe (changes requested):
I mistakenly left the above review as a `Comment`. I've now submitted it as a `Request changes`, so please take a look at that review! 🥲

src/main/kotlin/lotto/Money.kt
mariyakulikova: Thank you for pointing it out! I will refactor this to use a named constant instead of the magic number.

src/main/kotlin/lotto/Lotto.kt
mariyakulikova: During pair programming we figured out that Kotlin has getters under hood and removed all duplicates that we did. Unfortunately this one was left by mistake.

src/main/kotlin/lotto/Lotto.kt
mariyakulikova: I wasn’t totally sure, but I tried using a Set to improve performance. Let me know if there’s a better way!

src/main/kotlin/lotto/Calculator.kt
mariyakulikova: We weren't sure where the best place to calculate the rate and print the result was, so we made a separate object for the division to test it more easily. I will remove it during refactoring game result structure

src/main/kotlin/lotto/LottoGame.kt
mariyakulikova: We were thinking about two options: one with state and one where functions return values. But we didn’t realize that keeping state could make things more complex. Thanks for pointing that out!

src/main/kotlin/lotto/Rank.kt
mariyakulikova: I changed it to `when (matchCount)`. It really looks easier to read now. Appreciate the tip!

src/main/kotlin/view/ResultView.kt
mariyakulikova: That makes sense! I moved the calculation out, now the view layer focused on displaying results.
One question - is it fine to put the logic in `main`, or should I make a new class for it?

src/test/kotlin/lotto/LottoGameTest.kt
mariyakulikova: Thanks for the feedback! I didn't think about it from reader's perspective.  I refactored the code to use a data class instead of a Map.

src/main/kotlin/lotto/Lotto.kt
Wordbe: Wow, that’s exactly it. Great approach - thank you! 😊

src/main/kotlin/lotto/Lotto.kt
Wordbe: Ah, I see! Thanks for reflecting that 😄

src/main/kotlin/lotto/LottoGame.kt
Wordbe: Great. Thank you!

src/main/kotlin/view/ResultView.kt
Wordbe: You might want to consider which object would be most appropriate to take responsibility for the calculation logic.

Try thinking about it in terms of the Single Responsibility Principle (SRP) 😊

src/test/kotlin/lotto/LottoGameTest.kt
Wordbe: That's a great approach! Creating a new object with explicit field names has definitely made it easier to understand. Thank you! 😊

Wordbe (approved):
I know it wasn’t easy to incorporate the feedback, but I really appreciate you taking the time to do it carefully and leaving thoughtful comments.

Also, please add any new points that need to be addressed to the beginning of the next mission. I’ll go ahead and merge this PR here!

Let’s move on to the next mission — thank you! 😊

PR #14, opened 2025-07-03
[Lotto - step1] Submit saragrosser's mission

saragrosser:
Lotto Step 1

This PR is to show the implementation of a Lotto Machine that allows users to buy lotto tickets with a set price of 1000 KRW. The machine calculates the winning tickets matching the numbers and giving back a return rate to the user.

Features
-Implementing the mental model of Lotto generation 
-Implementing the possibility for user input 
-Implementing the rank for creating the winning ticket 
-Implementing the output of the machine based on the matching numbers 

Additional Comments: 
This project was created practicing the TDD method with my pair colleague, this was challenging at the beginning but with practice it will become better. We also created a design using the MVC method which allow us to work together as a team and that was extremely helpful . 

Feedback welcome!
I am really greatful for your honest feedback, please feel free to comment on the project, and I would like if possible to get a feedback on the Rank.kt object file to learn how to refactor it better 

Thank you so much! 
Sara


README.md
```diff
@@ -1 +1,61 @@
...
+- Get used to TDD.
+- Get used to Kotlin API and JUnit.
+- Get used to the pair programming.
+- Not get pressure on creating a perfect project, rather make mistake and learn.
+
+
+## Features List
+
+Lotto Machine
+- [x] e. Need input from user for the amount of money
```
Wordbe, line 19: What do 'e, em, m' means? 🙂

README.md
```diff
@@ -1 +1,61 @@
...
+- [x] m. calculate winning money
+- [ ] e. winning ticket(s) --> LottoMachine
+- [ ] m. calculate winning ticket(s) --> LottoMachine
+- [x] e. return rate
+- [x] m. calculate return rate
+
+Rank
+- [x] Rank is enum class
+- [x] e. count of match
+- [ ] count of match is magic number
```
Wordbe: If you compleyed, then check them 🙂

src/main/kotlin/lotto/Application.kt
```diff
@@ -0,0 +1,34 @@
+package lotto
+
+fun main() {
+    val amountOfMoney = retryable { InputView.getPurchaseAmount() }
```
Wordbe: we should consider whether Application.kt should be responsible for this feature.

src/main/kotlin/lotto/Application.kt
```diff
@@ -0,0 +1,34 @@
...
+    val lastWeekWinningNumbers = retryable { InputView.getLastWeekWinningNumbers() }
+    val bonusNumber = retryable { InputView.getBonusNumber() }
+    println()
+
+    val machine = LottoMachine(amountOfMoney, lastWeekWinningNumbers, bonusNumber)
+    machine.bundleOfLottoTicket = printer.bundleOfLottoTicket
+    machine.writeResultTable()
+
+    val winStats = machine.winStat
+    winStats.calculateWinningMoney(machine.resultTable)
```
Wordbe: Seeing `calculate~`...

Instead of storing the result in an internal variable and retrieving it later, there's a way to return the result immediately and pass it naturally within the block. 🙂

From a state management and side-effect perspective, I find the latter approach better. What are your thoughts?

src/main/kotlin/lotto/InputView.kt
```diff
@@ -0,0 +1,37 @@
+package lotto
+
+object InputView {
```
Wordbe, line 3: Let’s complete the TODO! ✅

src/main/kotlin/lotto/InputView.kt
```diff
@@ -0,0 +1,37 @@
...
+    // TODO: Have to deal with the exception
+    // How about dealing it with retryable command?
+    fun getPurchaseAmount(): Int {
+        println("Please enter the purchase amount.")
+        val input = readln()
+        val userInput = input.toIntOrNull() ?: throw IllegalArgumentException("This is not a number.")
+        require(userInput % 1000 == 0)
```
Wordbe: It would be great to understand the drawbacks of magic numbers and replace them with constants that have meaningful names. 😊

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,35 @@
+package lotto
+
+class LottoMachine(
+    val amountOfMoney: Int,
+    val lastWeekWinningNumbers: List<Int>,
+    val bonusNumber: Int,
+) {
+    val amountOfTicket: Int = amountOfMoney / LottoTicket.COST_OF_TICKET
```
Wordbe, line 8: (If time allows, later...)
How about avoiding internal variables beyond those in the primary constructor?
Returning results directly from methods each time can eliminate side effects, clarify the object’s responsibilities, and reduce complexity. 🙂

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,35 @@
...
+    lateinit var bundleOfLottoTicket: List<LottoTicket>
+    val resultTable: MutableList<Int> = MutableList(6) { 0 }
```
Wordbe: It would be better if resultTable had a more specific and descriptive name—right now, it’s unclear what kind of result it represents. 🙂
The meaning of each index could also be ambiguous.
How about expressing this as an object that holds both the rank and its corresponding count?

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,35 @@
...
+            Rank.FIRST -> resultTable[1]++
+            Rank.SECOND -> resultTable[2]++
+            Rank.THIRD -> resultTable[3]++
+            Rank.FOURTH -> resultTable[4]++
+            Rank.FIFTH -> resultTable[5]++
+            Rank.MISS -> resultTable[0]++
+        }
+    }
+
+    private fun getRankOfTicket(ticket: LottoTicket): Rank {
```
Wordbe: Private methods often reveal responsibilities that might actually belong to another object. 
How about moving this method to the Rank object instead?

src/main/kotlin/lotto/LottoPrinter.kt
```diff
@@ -0,0 +1,16 @@
+package lotto
+
+class LottoPrinter(val amountOfMoney: Int) {
+    val amountOfTicket: Int = amountOfMoney / LottoTicket.COST_OF_TICKET
+    val bundleOfLottoTicket: List<LottoTicket> = generateLottoTickets(amountOfTicket)
+
+    fun generateLottoTickets(amountOfTicket: Int): List<LottoTicket> {
```
Wordbe, line 7: Shall we try rewriting this in a functional style? 🙂
Functional programming offers many benefits—immutability, declarative code, and conciseness. Kotlin supports this style very well!

src/main/kotlin/lotto/LottoTicket.kt
```diff
@@ -0,0 +1,16 @@
+package lotto
+
+// TODO: Generated by RandomNumberGenerator
```
Wordbe: If this were production code, having such default values could be risky. ⚠️
Also, please make sure to remove the TODO!

src/main/kotlin/lotto/RandomNumberGenerator.kt
```diff
@@ -0,0 +1,9 @@
+package lotto
+
+object RandomNumberGenerator {
+    fun generateNumber(): List<Int> {
```
Wordbe, line 4: Great 👍

src/main/kotlin/lotto/Rank.kt
```diff
@@ -0,0 +1,30 @@
...
+) {
+    FIRST(6, false, 2_000_000_000),
+    SECOND(5, true, 30_000_000),
+    THIRD(5, false, 1_500_000),
+    FOURTH(4, false, 50_000),
+    FIFTH(3, false, 5_000),
+    MISS(0, false, 0), ;
+
+    companion object {
+        fun of(
```
Wordbe, line 17: I recommend when expresion like `when (matchCount)`🙂

src/test/kotlin/lotto/LottoMachineTest.kt
```diff
@@ -0,0 +1,3 @@
+package lotto
+
+class LottoMachineTest
```
Wordbe: Need to be tested. right?

Wordbe (changes requested):
Hello! saragrosser
Thank you for the thoughtful code and explanation. 😊

I’ve left some comments on areas that might be worth reconsidering—please take a look and make updates where needed.
Also, it’d be great to add tests for other classes like LottoPrinter and LottoMachine if possible.

Thanks again!! 🙌

src/main/kotlin/lotto/Application.kt
saragrosser: I moved all main application logic into a new Lotto object in Lotto.kt to improve code organization and readability. Now Application.kt only contains a single entry point.

README.md
saragrosser: Sorry I have now added acronym description to the readme file, to describe: e = entity, m = method, em = entity + method :)

README.md
saragrosser: I have made the changes to complete the check list, thank you!

src/main/kotlin/lotto/RandomNumberGenerator.kt
saragrosser: Thank you!

src/main/kotlin/lotto/Application.kt
saragrosser: I have refactored WinningStatistics logic into a stateless utility object WinStatCalculator as suggested. I removed internal state management from LottoMachine and Lotto.kt, replaced calculateWinningMoney and calculateReturnRate with pure functions as a result WinningStatistics.kt  was removed

Thanks for the feedback! It helped me simplify and improve the design. 😊

src/main/kotlin/lotto/InputView.kt
saragrosser: I’ve removed the comment from InputView.kt as it’s no longer needed. Because I was rushing to handle it over we forgot to clean the comments,  I will keep this in mind for the next missions!! Thank you

src/main/kotlin/lotto/InputView.kt
saragrosser: I changed out the  1000 for the shared LottoTicket.COST_OF_TICKET constant.  Is this a better approach? thank you!

src/main/kotlin/lotto/LottoMachine.kt
saragrosser: I updated resultTable from List<Int> to Map<Rank, Int> and adjusted the logic across  the related files :)

src/main/kotlin/lotto/LottoMachine.kt
saragrosser: Thank you this has help me understand better the SRP principle I have moved the function to the Rank object!

src/main/kotlin/lotto/LottoPrinter.kt
saragrosser: I’ve updated the function. I have to study more about functional programming, thank you! Let me know if there’s anything else I can improve!

saragrosser: I’ve updated the function. Let me know if there’s anything else I can improve! :)

src/main/kotlin/lotto/RandomNumberGenerator.kt
saragrosser: thank you

src/main/kotlin/lotto/LottoTicket.kt
saragrosser: Thank you I will keep these in mind for the future for now I’ve removed the default value to ensure tickets must be created explicitly. 
And also removed the TODO sorry about that!

src/main/kotlin/lotto/Rank.kt
saragrosser: Thank you I have used when (matchCount) now to refactor this

src/test/kotlin/lotto/LottoMachineTest.kt
saragrosser: Done, I added a test that verifies LottoMachine ranks tickets correctly

src/main/kotlin/lotto/Rank.kt
saragrosser: thank you I have change this

src/main/kotlin/lotto/LottoMachine.kt
saragrosser: Unfortunately I have very little span time, but I will keep this in mind for Step 2, thank you!

src/main/kotlin/lotto/InputView.kt
saragrosser: After investigating this, I learned that using magic numbers (hard-coded numeric values) makes code harder to read, understand, and maintain because the meaning behind the number isn't clear. If the same number is used in multiple places, updating it later becomes error-prone. Instead, using named constants improves readability by giving context, makes the code easier to change, and helps prevent bugs.

README.md
Wordbe: Thank you very much for your explanation 🙂

src/main/kotlin/lotto/Application.kt
Wordbe: Thank you for reflecting the feedback! 😊

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -0,0 +1,20 @@
+package lotto
+
+class LottoMachine(
```
Wordbe, line 3: Try removing any variables in the class that aren’t declared in the primary constructor.
For example, in LottoMachine, amountOfTicket can be turned into a method instead of being stored as a variable.
Likewise, bundleOfLottoTicket doesn’t really need to be stored as a field — you could pass it in as a parameter when needed.
As for resultTable, rather than keeping it as state, try having a method return it directly.

Holding internal state can make the code harder to understand, and it increases the chance of bugs due to side effects. Let’s aim for a more functional and predictable structure! 😊

src/main/kotlin/lotto/InputView.kt
Wordbe: Great job! thank you🫡

Wordbe (approved):
Hi Sari!

I’ve reviewed the updates you made based on my previous feedback — thank you for taking the time, and great job!

I’ll go ahead and approve and merge this PR to move on to the next step.

That said, I believe there’s still a lot to gain if you take a moment to think more deeply about and reflect on the following points in the early part of the next step:

Remove unnecessary internal state from classes like LottoMachine, LottoPrinter, and LottoTicket Strengthen test coverage for all core functionalities 

Thanks again for all your hard work — I truly appreciate the effort you put in! 😊

PR #15, opened 2025-07-03
[Lotto - step1] Submit vittorio's mission

VittorioDeMarzi:
We used the TDD and pair programming to complete the LOTTO project, following the Kotlin conventions.
We used the AngularJS commit conventions to write Git commits.
Writing first the tests and then the logic was challenging but thanks to it we learned a lot. We maybe need more time to refactor the project.

We would like to get feedbacks on:
 - how to write tests more efficiently
 - how to follow a clean architecture


README.md
```diff
@@ -1 +1,12 @@
...
+## Features
+
+- [ ] Purchase amount validation (Greater or equal than 1000, divisible by 1000)
+- [ ] Generate number of tickets
+- [ ] Generate LottoTickets
+  - each ticket has 6 numbers
+- [ ] Validation winning ticket
+- [ ] Validation Bonus number
+- [ ] Get winning prizes
+- [ ] Calculate return rate
```
Hyeon9mak: 🙄 are those all completed?

Organizing business or feature requirements in documents like the README.md(or something), will be helpful in the future whether it's you revisiting the project later on, or a fellow developer joining the team and looking at the project for the first time.

Hope you try this habit and keep refining it!

src/main/kotlin/lotto/ui/InputView.kt
```diff
@@ -0,0 +1,22 @@
+package lotto.ui
+
+object InputView {
+    fun purchaseAmountInput(): Int {
+        println("Please enter the purchase amount.")
+        val purchaseAmount =
+            readln().toIntOrNull() ?: throw IllegalArgumentException("Invalid input")
+        return purchaseAmount
+    }
```
Hyeon9mak: how about this? 😉 
it removes the `purchaseAmount`, avoids memory allocation.

```suggestion
        return readln().toIntOrNull() ?: throw IllegalArgumentException("Invalid input")
    }
```

src/main/kotlin/lotto/ui/InputView.kt
```diff
@@ -0,0 +1,22 @@
+package lotto.ui
+
+object InputView {
+    fun purchaseAmountInput(): Int {
+        println("Please enter the purchase amount.")
```
Hyeon9mak: it doesn't have any properties. it would be good to use the object class 👍 💯

src/main/kotlin/lotto/core/InputValidation.kt
```diff
@@ -0,0 +1,14 @@
...
+        try {
+            val winningLottoList = winningLottoString.split(",").map { it.toInt() }
+            require(winningLottoList.size == 6) { throw IllegalArgumentException() }
+
+            return winningLottoList
+        } catch (e: NumberFormatException) {
+            throw IllegalArgumentException("Invalid input")
+        }
+    }
+}
```
Hyeon9mak: What if we changed from console string input to web browser image selection?
Between `core` and `view`, whose responsibility would it be to split by `,`?

If a fellow developer gets a requirement saying "change the lotto number input to 5 numbers," 
which class file would they look at first - `Lotto` or `InputValidation`?

I believe that thinking about how to help others when they examine my code in the future is exactly the standard for choosing where to assign object responsibilities.

src/main/kotlin/lotto/core/Lotto.kt
```diff
@@ -0,0 +1,21 @@
+package lotto.core
+
+data class Lotto(private val ticket: Set<Int>) {
+    constructor(ticket: List<Int>) : this(ticket.toSet())
+
+    init {
+        require(ticket.size == 6)
+    }
```
Hyeon9mak: `ticket.size`? is that means 'lotto ticket paper size' or something....? 🤔 
i think that need more intuitive name! 😄

src/main/kotlin/lotto/core/Lotto.kt
```diff
@@ -0,0 +1,21 @@
+package lotto.core
+
+data class Lotto(private val ticket: Set<Int>) {
+    constructor(ticket: List<Int>) : this(ticket.toSet())
```
Hyeon9mak: Since duplicates aren’t allowed. Set is a nice choice! 👍 👍

src/main/kotlin/lotto/core/Lotto.kt
```diff
@@ -0,0 +1,21 @@
...
+        return this.ticket.intersect(winningTicket.ticket).size
+    }
+
+    fun containsBonus(bonusNumber: Int): Boolean {
+        return ticket.contains(bonusNumber)
+    }
+
+    fun displayLotto() {
+        println(ticket.joinToString(",", prefix = "[", postfix = "]"))
+    }
```
Hyeon9mak: What if we changed from console string input to web browser image selection?
It will be changed both `InputView`, `OutputView` and `Lotto`.

That would be annoying and bothersome. 
How about we leave the output methods completely to the view classes?

src/main/kotlin/lotto/core/LottoLogic.kt
```diff
@@ -0,0 +1,26 @@
...
+        repeat(numberOfTickets) {
+            val lotto = Lotto(generateListOfSixRandomNumbers())
+            listOfTickets.add(lotto)
+        }
+        return listOfTickets
+    }
+
+    fun generateWinningLotto(listOfInt: List<Int>): Lotto {
+        return Lotto(listOfInt)
+    }
```
Hyeon9mak: 😮 The name seems to have every logic of lotto.
I think it will be grow up 1_000_000 lines super class.

Since it's so common, we'll probably need to open this class file every time requirements change.
That would increase the risk of merge conflicts when working with fellow developers.

How about separating the logic - asking `LottoTicket` should go to the `LottoTicket` class, and `WinningLotto` generation should go to another class?
We could divide them by name and create separate classes.

src/main/kotlin/lotto/core/Rank.kt
```diff
@@ -0,0 +1,31 @@
...
+            return when {
+                countOfMatch == FIRST.countOfMatch -> FIRST
+                countOfMatch == SECOND.countOfMatch && matchBonus -> SECOND
+                countOfMatch == THIRD.countOfMatch -> THIRD
+                countOfMatch == FOURTH.countOfMatch -> FOURTH
+                countOfMatch == FIFTH.countOfMatch -> FIFTH
+                else -> MISS
+            }
+        }
+    }
```
Hyeon9mak: nice! if winningMoney changes or new rank are added, only need to look at the Rank enum class! 💯 👍
how about append property of matchBonus?
it will be express the condition of difference between SECOND rank and THRID rank!

src/main/kotlin/lotto/core/WinningLogic.kt
```diff
@@ -0,0 +1,29 @@
...
+//        tickets.forEach {
+//            val countMatches = it.countMatches(winningLotto)
+//            val hasBonus = it.containsBonus(bonusNumber)
+//            val rank = lotto.core.Rank.valueOf(countMatches, hasBonus)
+//            println(rank)
+//            mapOfResult.put(rank, mapOfResult.getOrDefault(rank, 0) + 1)
+//        }
+
+//        return mapOfResult
+    }
```
Hyeon9mak: Plz remove all unused comments.
Fellow developers might read the comments and wonder 'Should I refer to this?' which could cause confusion and reduce productivity. 😢

src/test/kotlin/lotto/core/InputValidationTest.kt
```diff
@@ -0,0 +1,22 @@
...
+        val winningLotto = inputValidation.createWinningLottoList(winningLottoString)
+        assertThat(winningLotto).hasSize(6)
+    }
+
+    @Test
+    fun `winning numbers should has to be exactly 6 numbers`() {
+        val inputValidation = InputValidation()
+        val winningLottoString = "1,2,3,4,5"
+        assertThrows<IllegalArgumentException> { inputValidation.createWinningLottoList(winningLottoString) }
+    }
```
Hyeon9mak: how about append more test case?! `1, 2, 3, 4, 5, 6, 7` 😉

src/test/kotlin/lotto/core/LottoLogicTest.kt
```diff
@@ -0,0 +1,38 @@
...
+class LottoLogicTest {
+    val lottoLogic = LottoLogic()
+
+    @ParameterizedTest
+    @ValueSource(ints = [1000, 2000, 3000, 4000])
+    fun `calculate the right number of tickets`(purchaseAmounts: Int) {
+        val rightNumberOfTickets = purchaseAmounts / 1000
+        val numberOfTickets = lottoLogic.calculateNumberOfTickets(purchaseAmounts)
+        assertThat(rightNumberOfTickets).isEqualTo(rightNumberOfTickets)
+    }
```
Hyeon9mak: What happens when negative, 0, 999 are passed? 🤔

src/test/kotlin/lotto/core/RankTest.kt
```diff
@@ -0,0 +1,46 @@
...
+        assertThat(rank).isEqualTo(Rank.FOURTH)
+    }
+
+    @Test
+    fun `value of fifth rank`() {
+        val matchCount = 3
+        val matchBonus = false
+        val rank = Rank.valueOf(matchCount, matchBonus)
+        assertThat(rank).isEqualTo(Rank.FIFTH)
+    }
```
Hyeon9mak, line 48: Very simple and good tests! 👍 
and.. What happens when 2 or less count passed?

src/main/kotlin/lotto/Application.kt
```diff
@@ -0,0 +1,32 @@
...
+import lotto.ui.ResultView
+
+fun main() {
+    val purchaseAmount = InputView.purchaseAmountInput()
+    val lottoMachine = LottoLogic()
+
+    val numberOfTickets = lottoMachine.calculateNumberOfTickets(purchaseAmount)
+    val tickets = lottoMachine.generateTickets(numberOfTickets)
+
+    val winningLottoString = InputView.winningNumbersInput()
```
Hyeon9mak: What happens when winning numbers other than 1-45 are passed?

Although it wasn't specifically mentioned in the requirements, 
I think developers who care about creating better applications for the business should pay attention to even these small details.

How about referring to the actual Korean lotto game rules?
https://en.lottolyzer.com/how-to-play/south-korea/6_slash_45-lotto

src/main/kotlin/lotto/Application.kt
```diff
@@ -0,0 +1,32 @@
...
+    val inputValidation = InputValidation()
+    val winningLottoList = inputValidation.createWinningLottoList(winningLottoString)
+    val winningLotto = lottoMachine.generateWinningLotto(winningLottoList)
+
+    val bonusNumber = InputView.bonusNumberInput()
```
Hyeon9mak: and the bonusNumber is same as lotto winning number.
it will generate duplicate validation code.

i think they can be wrapped a VO(Value Object) or something and put validation code in it... 😉

Hyeon9mak (changes requested):
Hi Vitto!
You did a great job to solve the problem and complete mission 👍 
Below are the answers to your questions.

#### how to write tests more efficiently

The most important reason for TestCode is to clear the "unstable parts" of your the program before you run it yourself.
What is ur unstable parts of this mission?

- lotto number range
- lotto number count
- no allow duplicate numbers
- and more something...

Take your time to think about what you need functionally and commercially.
Think about the rules that you must take in the real world.
I believe you can easily find out what to test 😄 

#### how to follow a clean architecture

What is the clean architecture? And what are its key principles?
As u know, the right answer varies depending on each situation.

I believe the clean architecture should be driven by business, not technology.
The architecture should be flexible enough to adapt to the business problems and characteristics that Vito is trying to solve.

I'll propose two situations.

- What kind of architecture would be good for a business where Views change frequently but Domain changes are rare?
- What kind of architecture would be good for a business where Views don't change but Domain changes are frequent?

With this kind of business context in mind, you can choose better way to follow the clean architecture.
How about setting the standard as "a business where Views change frequently but Domain changes are rare" during the HeroTechCourse mission? 😉 

Feel free to ask any questions you have!

README.md
VittorioDeMarzi: You're totally right.
We started coding a bit too quickly because we were worried about the deadline, and we didn’t organize the README well.
Having a clear and updated README is really helpful, both for planning and for anyone who looks at the project.

src/main/kotlin/lotto/ui/InputView.kt
VittorioDeMarzi: It makes sense thanks. I’ll update it like this.

src/main/kotlin/lotto/core/InputValidation.kt
VittorioDeMarzi: Thanks, I really liked your point about responsibility.
I updated the code: now InputView creates the list of numbers from the string, and InputValidation just checks the list.
I hope this is the right approach and would be happy to get more feedback on it! :)

src/test/kotlin/lotto/core/InputValidationTest.kt
VittorioDeMarzi: Great point! 😊
I added the test case with 1,2,3,4,5,6,7 to cover inputs with too many numbers.

src/main/kotlin/lotto/core/Lotto.kt
VittorioDeMarzi: It sounds confusing 😄 I renamed it to something more clear, like numbers.

src/main/kotlin/lotto/core/Lotto.kt
VittorioDeMarzi: Thanks, I really appreciate this feedback :)

src/main/kotlin/lotto/core/Lotto.kt
VittorioDeMarzi: It makes more sense to leave output to the view layer.
I moved the display logic to ResultOutput now 😊

src/main/kotlin/lotto/core/Rank.kt
VittorioDeMarzi: Thanks to this advice I have even changed the logic of valueOf.
The matchBonus condition is now integrated directly in the valueOf logic, improving readability :)

src/test/kotlin/lotto/core/LottoLogicTest.kt
VittorioDeMarzi: Good question :D

I had the requirements in the InputValidation

`fun validatePurchaseAmount(input: String): Int {`
`      val purchaseAmount = input.toIntOrNull() ?: throw`
`      IllegalArgumentException(Error.INVALID_INPUT.message)`
`      require(purchaseAmount >= 1000) { Error.LOW_PURCHASE_AMOUNT.message }`
`      require(purchaseAmount % 1000 == 0) { Error.NOT_DIVISIBLE_BY_1000.message }`
`      return purchaseAmount`
`} `

should I add it even in this method?

src/main/kotlin/lotto/Application.kt
VittorioDeMarzi: Totally agree, under pressure we ended up making too many assumptions and skipped over important validations like this one.
We're definitely learning from these mistakes, and your feedback is super helpful to keep me on the right track.

src/main/kotlin/lotto/Application.kt
VittorioDeMarzi: I've now implemented a `WinningLotto` class as a Value Object that wraps both the winning numbers and the bonus number.

The validation logic, including preventing the bonus number from being part of the winning numbers, now lives inside that class.

I have a question 🙂
When I was learning Java, I was told it's a good practice to validate critical inputs in multiple layers, for example, both in the controller and in the domain/model, to increase robustness and avoid relying on a single point of failure.

Now I see that repeating the same validation logic in different places can lead to code duplication and inconsistency. From your comment, I understand that wrapping the lotto numbers and bonus number into a Value Object could help centralize the validation logic.

Would you recommend moving all validation into the Value Object (like Lotto or WinningLotto), and removing validation from elsewhere (e.g. InputView)? Or is some redundancy still acceptable in certain layers?

I'm still learning, so I'd love to hear your take on this balance between safety and clean design 😊

src/main/kotlin/lotto/core/LottoLogic.kt
VittorioDeMarzi: To address this, I already refactored the logic into two separate components:
- TicketFactory: handles ticket generation and the purchase amount logic.
- WinningLottoFactory: creates the WinningLotto from the main numbers and the bonus number.

README.md
```diff
@@ -1 +1,23 @@
...
+
+## Features
+
+- [x] Purchase amount validation (Greater or equal than 1000, divisible by 1000)
+    - [x] The input amount must be greater than or equal to 1,000 KRW
+    - [x] The amount must be divisible by 1,000 KRW
+- [x] Calculate number of tickets from the amount: amount / 1000
+- [x] Generate LottoTickets
+    - [x] each ticket has 6 numbers
+    - [x] Numbers must be in the range 1 to 45
```
Hyeon9mak, line 11: 😉 👍 💯

src/main/kotlin/lotto/model/InputValidation.kt
```diff
@@ -0,0 +1,26 @@
+package lotto.model
+
+import lotto.view.Error
+
+object InputValidation {
+    fun validatePurchaseAmount(input: String): Int {
+        val purchaseAmount = input.toIntOrNull() ?: throw IllegalArgumentException(Error.INVALID_INPUT.message)
```
Hyeon9mak, line 7: like the `validateBonusNumberInput`, purchaseAmount input parameter can also use int.

src/main/kotlin/lotto/model/InputValidation.kt
```diff
@@ -0,0 +1,26 @@
...
+        require(numbers.all { it in 1..50 }) { Error.OUT_OF_RANGE.message } // any number must be in range 1 - 50
+        require(numbers.size == 6) { Error.INVALID_LOTTO_SIZE.message } // Winning lotto has to be 6 numbers
+    }
+
+    fun validateBonusNumberInput(
+        winningNumbers: List<Int>,
+        bonusNumberInput: Int,
+    ) {
+        require(bonusNumberInput in 1..50) { Error.OUT_OF_RANGE.message } // bonus number must be in range 1 - 50
+        require(!winningNumbers.contains(bonusNumberInput)) { Error.BONUS.message } // bonus number must not be in the winningLotto numbers
```
Hyeon9mak, line 24: I think the current code is a great approach!
If the policies for `winningNumber`, `bonusNumber`, and `purchase` change independently in different scenarios, the current code could become a maintenance burden.
But if the entire Lotto game rules tend to change frequently all together, then the current code would actually be excellent for maintenance.

As I mentioned before, whether the current code is beneficial or detrimental for maintenance depends on the business characteristics.

On a separate note, since you've experimented with this approach, I'd recommend trying missions where policies are separated by domain models as well.
I believe you need to experience both sides to understand their respective advantages and disadvantages, which will help you make better decisions quickly when new requirements come up.

src/main/kotlin/lotto/model/Lotto.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import lotto.view.Error
+
+data class Lotto(private val numbers: Set<Int>) {
+    constructor(numbers: List<Int>) : this(numbers.toSet())
+
+    init {
+        require(numbers.all { it in 1..50 }) { Error.OUT_OF_RANGE.message } // any number must be in range 1 - 50
+        require(numbers.size == 6) { Error.INVALID_LOTTO_SIZE.message } // Lotto has to be 6 numbers
+    }
```
Hyeon9mak, line 11: Duplicate validation with `InputValidation`.
If the policies changed, fellow developer need to check all class files.

How about unify validation logic separate or combine?

src/main/kotlin/lotto/model/Lotto.kt
```diff
@@ -0,0 +1,24 @@
...
+        return this.numbers.intersect(winningTicket.numbers).size
+    }
+
+    fun containsBonus(bonusNumber: Int): Boolean {
+        return numbers.contains(bonusNumber)
+    }
+
+    fun displayLotto(): String {
+        return numbers.joinToString(",", prefix = "[", postfix = "]")
+    }
```
Hyeon9mak, line 23: There's still a bit of string output responsibility left. 😉 
how about we leave this to the view too?

src/main/kotlin/lotto/model/TicketFactory.kt
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    fun generateTickets(numberOfTickets: Int): List<Lotto> {
+        val listOfTickets = mutableListOf<Lotto>()
+        repeat(numberOfTickets) {
+            val lotto = Lotto(Random.generateListOfSixRandomNumbers())
+            listOfTickets.add(lotto)
+        }
+        return listOfTickets
+    }
```
Hyeon9mak, line 19: In the future, when we want to modify Ticket-related policies, we can just look for `TicketFactory` 😉 👍 
(and there's duplicated validation logic here too! 🥲)

src/main/kotlin/lotto/model/WinningLogic.kt
```diff
@@ -0,0 +1,16 @@
...
+) {
+    fun determineWinningTickets(): Map<Rank, Int> {
+        return tickets
+            .map {
+                val countMatches = winningLotto.countMatches(it)
+                val hasBonus = winningLotto.isBonusMatched(it)
+                Rank.valueOf(countMatches, hasBonus)
+            }.groupingBy { it }
+            .eachCount()
+    }
```
Hyeon9mak, line 15: I think... it can be changed to non-property class.
What do u think? what are the pros & cons? 🤔

src/main/kotlin/lotto/model/Rank.kt
```diff
@@ -0,0 +1,26 @@
...
+    companion object {
+        fun valueOf(
+            countOfMatch: Int,
+            matchBonus: Boolean,
+        ): Rank {
+            return entries.find {
+                it.countOfMatch == countOfMatch && it.matchBonus == matchBonus
+            } ?: MISS
+        }
+    }
```
Hyeon9mak, line 21: Nice! 💯

src/main/kotlin/lotto/model/Rank.kt
```diff
@@ -0,0 +1,26 @@
+package lotto.model
+
+enum class Rank(val countOfMatch: Int, val winningMoney: Int, val matchBonus: Boolean = false) {
+    FIRST(6, 2_000_000_000),
+    SECOND(5, 30_000_000, matchBonus = true),
+    THIRD(5, 1_500_000),
+    FOURTH(4, 50_000),
+    FIFTH(3, 5_000),
+    MISS(0, 0),
+    ;
```
Hyeon9mak, line 10: Good to see the default parameter value 👍 💯

src/main/kotlin/lotto/model/WinningLotto.kt
```diff
@@ -0,0 +1,16 @@
...
+    init {
+        require(!lotto.containsBonus(bonusNumber)) { throw IllegalArgumentException(Error.BONUS.message) }
+        require(bonusNumber in 1..45) { Error.BONUS_OUT_OF_RANGE }
+    }
+
+    fun countMatches(other: Lotto): Int {
+        return lotto.countMatches(other)
+    }
+
+    fun isBonusMatched(lotto: Lotto): Boolean = lotto.containsBonus(bonusNumber)
```
Hyeon9mak, line 15: data class! what is the different with class, and value class?

src/test/kotlin/lotto/core/LottoLogicTest.kt
Hyeon9mak: Sure 😉

src/main/kotlin/lotto/Application.kt
Hyeon9mak: I think this is such a great question! 👍 

To give you the conclusion first,
I'd like you to experience all of these approaches.
I believe that these accumulated experiences will help you design better code structures more quickly.

If validation logic exists duplicated across multiple layers (view, domain), data reliability would increase, but in an environment where policies change frequently, there would be more code to modify in multiple places, which creates a risk that someone might accidentally miss a change.

So as I mentioned earlier, depending on business characteristics, you could either gather policies in one class file or distribute validation logic so that each domain has its own policies.

As you well know, there's no silver bullet. 
Ultimately, HeroTechCourse is time to experience all of this and grow, so I hope you take your time and enjoy the process. 😄

Hyeon9mak (approved):
Hi Vito!
Great job the reflect the feedback!

I've added a few more comments, but you seem to have a good grasp of everything and only need minor code tweaks. 😄 

so I'll advance you directly to the next stage. See you in step 2!
If you have more questions, plz feel free to leave a comment 😉

src/main/kotlin/lotto/model/WinningLotto.kt
VittorioDeMarzi: A data class is used when the class is only meant to hold data
Kotlin automatically generates useful functions for you:
 - equals() and hashCode() → so you can compare objects
 - toString() → for readable printing
 
 A regular class in Kotlin is the most flexible type. You use it when your object has both data and behavior and nothing is generated for you.
 
 A value class wraps exactly one property, typically used for creating a strong type for something like Email, UserId, or MoneyAmount. (Only one property (val) allowed)

src/main/kotlin/lotto/model/WinningLogic.kt
VittorioDeMarzi: Turning WinningLogic into an object makes sense if it has no state and just provides pure behavior. I initially had it as a class because I was passing in tickets and winningLotto as dependencies, which made it easier to test and reuse with different inputs.

However, since the logic inside determineWinningTickets() is pure and doesn’t rely on internal state across multiple method calls, I refactored it as an object and made determineWinningTickets() a function that receives the inputs directly.

Pros of using object:
 - No need to instantiate — it's stateless and can be reused globally.
 - Clearer that it's just utility logic.
 - Slightly simpler code.

Cons:
 - Less flexible if we later want to inject different strategies or behaviors.

src/main/kotlin/lotto/model/InputValidation.kt
VittorioDeMarzi: Initially, I thought having validation in two layers was a better practice, one in InputValidation to catch errors early in the input phase, and another in the domain classes (like Lotto) to guarantee object integrity regardless of where or how they're created.

My idea was to fail fast on user input, but also make sure invalid domain objects can never exist, even if created programmatically elsewhere (e.g. in tests, refactors, or future business logic).

But now I understand your point: if policies change frequently together, then centralizing validation might actually be more maintainable. I really appreciate the suggestion of trying both approaches, that makes total sense, and I’ll definitely experiment with separating validation per model in the next mission.

Thanks again, I'm learning a lot from this process! 🙏

PR #17, opened 2025-07-07
[Lotto - Step2] Submit Apoorva's mission

Apoorva-S-Natesh:
This pull request adds a feature to enter tickets manually to the console-based lotto. It covers the core game logic, input and output handling, and error management.
Manual ticket count is validated, and a Ticket is generated for every manual ticket.
I would like feedback and suggestions on:

* Improvement on best practices of Kotlin
* Improvement on Design Pattern
* Improvement for functions designing
* Improvement for test cases
* File structure


README.md
```diff
@@ -39,4 +39,9 @@ OutputView
...
 Logic
 - [x] Calculate Winning Stats
-- [x] Calculate Return Rate
\ No newline at end of file
+- [x] Calculate Return Rate
+
+Manual Ticket
+- [x] number of manual tickets > possible ticket purchase
+- [x] manual tickets 0 or less than possible ticket purchase
+- [x] manual tickets are Tickets (check for being a valid ticket)
```
pkeugine: Again, nice work on README 👍 
Also, the commit log looks even better than before 💯

src/main/kotlin/lotto/model/LottoNumber.kt
```diff
@@ -23,4 +27,4 @@ class LottoNumber private constructor(private val value: Int) {
     override fun toString(): String = value.toString()
 
     fun toInt(): Int = value
-}
+}
```
pkeugine: Screenshot 2025-07-07 at 11 24 04 PM

I see a red indicator at the end of some files.
I'd like you to find out what this sign means, and why we should avoid it.

**Hint**: [it's related to missing empty lines](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file) 😉

In IntelliJ, there's an option to automatically add a newline at the end of a file. It’s worth checking that out.

src/main/kotlin/lotto/model/LottoNumber.kt
```diff
@@ -12,6 +12,10 @@ class LottoNumber private constructor(private val value: Int) {
         }
 
         fun allNumbers(): List<LottoNumber> = NUMBERS.values.toList()
+
+        fun takeRandom(count: Int): List<LottoNumber> {
+            return (allNumbers()).shuffled().take(count).sortedBy { it.toInt() }
+        }
```
pkeugine: Since `allNumbers` method is only used in internally, why don't we make it private?

src/main/kotlin/lotto/model/LottoMachine.kt
```diff
@@ -0,0 +1,122 @@
...
+        repeat(ticketCount) {
+            val ticket = Ticket(generateTicketNumbers())
+            ticketsList.add(ticket)
+        }
+        OutputView.displayTickets(ticketCount, ticketsList, manualTicketCount)
+        return ticketsList
+    }
+
+    private fun generateTicketNumbers(): List<LottoNumber> {
+        return LottoNumber.Companion.takeRandom(Constants.NUMBER_COUNT)
```
pkeugine:
```suggestion
        return LottoNumber.takeRandom(Constants.NUMBER_COUNT)
```

`Companion` keyword isn't necessary when we call methods or use the variables in it.

src/main/kotlin/lotto/model/LottoMachine.kt
```diff
@@ -0,0 +1,122 @@
...
+    fun start() {
+        val amount = getAmountInput()
+        val manualTicketCount = getManualTicketCount(amount)
+        val manualTickets = getManualTickets(manualTicketCount)
+        val automaticTickets = generateAutomaticTickets(amount, manualTicketCount)
+        val winningNumbers = InputView.inputWinningNumbers()
+        val bonusNumber = InputView.inputBonusNumber(winningNumbers)
+        calculateStats(automaticTickets + manualTickets, winningNumbers, bonusNumber)
+        OutputView.displayResults(winStats, calculateReturnRate(amount))
+    }
```
pkeugine: I see you've moved `LottoMachine` into `model` directory.
When we talk about MVC pattern, controller's role is to connect the model and view so that our application gets shown as we expect it to. When I look at `LottoMachine`'s logic, it really feels like this is the `controller` in MVC; using domains and calling `InputView` & `OutputView`.

How do you think?
I'd like to know the thoughts behind this architecture :)

src/main/kotlin/lotto/model/LottoMachine.kt
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+    }
+
+    private fun getManualTickets(manualTicketCount: Int): List<Ticket> {
+        while (true) {
+            try {
+                val manualTicketsString = InputView.inputManualTicketNumbers(manualTicketCount)
+                val manualTickets = ManualTicket.generateManualTickets(manualTicketsString)
+                return manualTickets
```
pkeugine:
```suggestion
                return ManualTicket.generateManualTickets(manualTicketsString)
```

When we return the value of a variable right after initializing it, we can skip the initialization part and return it right away.

src/main/kotlin/lotto/model/LottoMachine.kt
```diff
@@ -0,0 +1,122 @@
...
+
+    private val winStats =
+        mutableMapOf(
+            Rank.FIRST to 0,
+            Rank.SECOND to 0,
+            Rank.THIRD to 0,
+            Rank.FOURTH to 0,
+            Rank.FIFTH to 0,
+            Rank.MISS to 0,
+        )
```
pkeugine: It'll be great if we could organize our class structure so that fields and methods are grouped together and be in an ordered place. You've done a great job of ordering the methods so let's apply that to non-method components as well.

src/main/kotlin/lotto/model/LottoMachine.kt
```diff
@@ -0,0 +1,122 @@
...
+
+    private fun calculateStats(
+        tickets: List<Ticket>,
+        winningNumbers: List<LottoNumber>,
+        bonusNumber: LottoNumber,
+    ) {
+        for (ticket in tickets) {
+            val match = checkMatch(ticket, winningNumbers)
+            if (match == 5 && ticket.numbers.contains(bonusNumber)) {
+                winStats[Rank.SECOND] = winStats[Rank.SECOND]!! + 1
```
pkeugine: I saw this `!!` operator in the previous step, but I decided not to comment on it yet.
Now, I think it's a good time for us to think about it :)

Do you know what this operator does?
I wouldn't say it's wrong, but I wanted to know if it were your intention.
Also, what are the alternatives to this approach?

src/test/kotlin/lotto/ManualTicketCountValidatorTest.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.junit.jupiter.api.assertThrows
+
+class ManualTicketCountValidatorTest {
+    @Test
+    fun `manual ticket count is a greater than possible tickets throws exception`() {
+        val exception =
+            assertThrows<IllegalArgumentException> {
+                ManualTicket.checkTicketCount(4, 2200)
+            }
+        assertThat("Manual tickets cannot be more than the possible purchase").isEqualTo(exception.message)
```
pkeugine:
```suggestion
        assertThat(exception).hasMessage("Manual tickets cannot be more than the possible purchase")
```

There's a nice tool in AssertJ for this situation :)

src/test/kotlin/lotto/ManualTicketCountValidatorTest.kt
```diff
@@ -0,0 +1,25 @@
+package lotto
+
+import lotto.model.ManualTicket
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+import org.junit.jupiter.api.assertThrows
+
+class ManualTicketCountValidatorTest {
```
pkeugine:
```suggestion
class ManualTicketTest {
```
When refactoring the application logic, don't forget to apply the changes to tests as well!

src/main/kotlin/lotto/model/LottoMachine.kt
pkeugine: This review is optional since it might take some time.

You did a great job creating `LottoMachine`.
Here's my thought on how to make it better.
I see that it is not only connecting the model and view together,
but also calculating stats and checking if ticket has the winning numbers.

In SRP point of view, I think we can separate those logic into other classes, either existing ones or new ones. For example, `checkMatch(ticket, winningNumbers)` looks like `Ticket` itself can handle it if we give winning numbers in method parameter.

By doing this, we'll also be able to create more tests that pin point on that particular logic. Currently, I feel like this `LottoMachine` holds important logic that aren't tested.

pkeugine (changes requested):
Hello Apoorva 👋 
I see you've made a very clean PR with improved commit logs 💯 

I left some comments, and you'll be able to handle most them quickly.
If some comments feel like they require a lot of work,
feel free to skip it for now and focus on deadline + next mission.

I'm open to any opinions or questions, so feel free to leave a comment or DM.
And as always,
Happy hacking 🚗 🚗 🚗

src/main/kotlin/lotto/model/LottoNumber.kt
Apoorva-S-Natesh: Changed the Settings on IntelliJ. "Ensure every saved file ends with a line break was not checked."

I had the habit of not adding a new line at the end of file (while working with C) , but I understand the importance of it now.

src/main/kotlin/lotto/model/LottoNumber.kt
Apoorva-S-Natesh: I agree, change Implemented

src/main/kotlin/lotto/model/LottoNumber.kt
Apoorva-S-Natesh: also I realized that while using version control systems like Git, not having a newline at the end of a file can cause unnecessary diffs. Adding or removing a newline might show up as a change, even though the actual content of the file hasn't been modified.

src/main/kotlin/lotto/model/LottoMachine.kt
Apoorva-S-Natesh: Thank you for catching that! Is Companion object like static so Members inside the companion object can be accessed using the class name?

src/main/kotlin/lotto/model/LottoMachine.kt
Apoorva-S-Natesh: I am confused between model and Controller. I thought model directly manages the data and the logic and rules of the application and is responsible for managing data of the app. So thought LottoMachine is a model since it handles data and manages the logic and rules of the application. 

But after reading a bit more, I understand what you are trying to convey. LottoMachine can be a controller since it controls all the task/events that user performs, navigation and Communication between model & view.

src/main/kotlin/lotto/model/LottoMachine.kt
Apoorva-S-Natesh: Thats a nice simplification. Thank you for the suggestion, I had completely missed that. I'll update the code.

src/main/kotlin/lotto/model/LottoMachine.kt
Apoorva-S-Natesh: change Implemented

src/main/kotlin/lotto/model/LottoMachine.kt
Apoorva-S-Natesh: I refactored the class and used Interface to separate concerns and responsibility. would it be better to add the statistic calculation and result calculation in Lotto Machine rather than Ticket, I feel Ticket is not responsible for calculating matches.

src/test/kotlin/lotto/ManualTicketCountValidatorTest.kt
Apoorva-S-Natesh: Thank you that's a very useful function!

src/test/kotlin/lotto/ManualTicketCountValidatorTest.kt
Apoorva-S-Natesh: Totally missed that!

src/main/kotlin/lotto/model/LottoMachine.kt
Apoorva-S-Natesh: There was a complitaion error "Operator call corresponds to a dot-qualified call 'winStats[rank].plus(1)' which is not allowed on a nullable receiver 'winStats[rank]' " So we deiced to use a Not Null Assertion Operator (!!) we used it as we were certain that the value will not be nul, it converts any nullable type to a non-nullable type and throws a Null Pointer Exception if the value is null.
on byte code it looks like 
`winStats.put(rank, ((Number)winStats.get(rank)).intValue() + 1);` and its using .intValue() which cannot receive null value before trying to increment it.

another safer option would be to use Elvis operator (?:) or getOrDefault 
`winStats[rank] = winStats[rank] ?.plus(1) ?: 1`
or
`winStats.put(rank, winStats.getOrDefault(rank, 0) + 1)`

src/main/kotlin/lotto/model/LottoMachine.kt
pkeugine: Yes, we can access them using the class name,
and kinda-yes about it being `like static`. (This is where `@JvmStatic` comes in)
I recommend [checking this out](https://kotlinlang.org/docs/object-declarations.html#companion-objects). It explains what companion object is in detail, with easy-to-read examples 👍

src/main/kotlin/lotto/model/LottoMachine.kt
pkeugine: Although I wrote a long-essay(?) about it, my thinking process was simple :  
-> This class has `InputView`, `OutputView`, and domains? Controller.
So yes, you perfectly understood what I tried to say. 🙏 

As for statistic / result calculation, I respect your opinion.
I suggest, and you decide 👍

src/main/kotlin/lotto/model/LottoMachine.kt
pkeugine: Nice!
To simplify, we get to choose if we want to throw the exception or to go on with our logic,
even though, like you said, we're pretty certain that it won't be null. 🤔 
It's just one of those 'machines don't understand us' moments, and I thought of another option :  
```Kotlin
winStats[Rank.SECOND] = winStats.getOrDefault(Rank.SECOND, 0) + 1
```

Your alternatives also work well, and even keeping `!!` isn't a wrong choice.
Nice explanation 👍

README.md
```diff
@@ -39,4 +39,9 @@ OutputView
 
 Logic
 - [x] Calculate Winning Stats
```
pkeugine, line 41: I actually try to go through the PR looking at individual commits,
and I see [a commit that can have a better title](https://github.com/herotechcourse/kotlin-lotto/pull/17/commits/27c46f99609a8faa69042bc4fc930907c17f5a6e) :)
You don't need to change this now, I just wanted to say that
it'll be good to try out 'changing the commit message and rebasing' when you have the time.

src/main/kotlin/lotto/service/Rank.kt
```diff
@@ -24,4 +24,4 @@ enum class Rank(val prize: Int) {
             }
         }
     }
-}
+}
```
pkeugine, line 27: Screenshot 2025-07-09 at 1 10 06 AM

I still see files with red indicator, so even if you checked the IntelliJ option, please make sure to apply it to files!

pkeugine (approved):
Hello Apoorva,
Nice job refactoring and checking all the reviews 💯 
I think we covered enough (more than enough I think) topics
and this was just our first mission!

You did a great job, and I'll merge this PR.
I left some additional comments, but they are very minor.

It was fun going through this mission with you, and good luck with Blackjack mission!
Happy hacking ♦️ ♦️ ♦️

README.md
Apoorva-S-Natesh: Noted, Will definitely try that! Thank you

src/main/kotlin/lotto/service/Rank.kt
Apoorva-S-Natesh: Thanks for pointing this out! I'll re-check the IntelliJ setting and apply it to all files.

PR #3, opened 2025-07-03
[Lotto - Step 1] Submit mehrshahbaz's mission

MehrShahbaz:
This pull request adds a console-based lotto game. It covers the core game logic, input and output handling, and error management.

I would like feedback on the following questions:

- File structure
- Improvement on best practices of Kotlin
- Improvement on Design Pattern
- Improvement for test cases
- Improvement for functions designing

PR #18, opened 2025-07-07
[Lotto - Step 2] Submit mehrshahbaz's mission

MehrShahbaz:
- How can Improve my test cases
- Is there any class that can be wrapped or created into another class ?
- When do I know when I have done too much nesting of class e.g
```
calculator.calculateReturnRate(machine.ticketCounter.purchaseAmount)
```

- is the use of data and value class to the standards ?


README.md
```diff
@@ -5,13 +5,22 @@
 - [x] Numbers should be between 0 and 50
 - [x] List has unique elements
 
+### Purchase Amount
+- [x] Minimum 1_000, (maximum 20_000 KRW)
+
+### Entered Ticket Count
+- [x] Must be a positive number
```
Hyeon9mak, line 12: detail policy 👍

src/main/kotlin/lotto/EnteredTicketCount.kt
```diff
@@ -0,0 +1,8 @@
+package lotto
+
+@JvmInline
+value class EnteredTicketCount(val count: Int) {
+    init {
+        require(count >= 0) { "Manual ticket count should be a positive number" }
+    }
+}
```
Hyeon9mak, line 8: VO makes easy to change data type when policy changed. 👍 
(If there was no wrapping with VO, fellow develop had to search and change every class file.)

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -2,12 +2,8 @@ package lotto
...
 class Lotto(val numbers: List<LottoNumber>) {
     init {
-        require(numbers.size == LOTTO_SIZE)
-        require(numbers.distinct().size == LOTTO_SIZE)
-    }
-
-    fun sortedList(): List<Int> {
-        return numbers.map { it -> it.value }.sorted()
+        require(numbers.size == LOTTO_SIZE) { "List should have exactly 6 numbers" }
+        require(numbers.distinct().size == LOTTO_SIZE) { "All the numbers in the list must be unique" }
```
Hyeon9mak: 😉 👍

src/main/kotlin/lotto/LottoHandler.kt
```diff
@@ -18,45 +33,51 @@ object LottoHandler {
...
+        }
+    }
+
+    private fun readManualTickets(count: Int): List<Lotto> {
+        val tickets = mutableListOf<Lotto>()
+        repeat(count) {
+            tickets.add(readTicket("Enter the numbers for manual tickets."))
+        }
+        return tickets
     }
```
Hyeon9mak: https://github.com/herotechcourse/kotlin-lotto/pull/3#discussion_r2184265927

> when switching from console screen to web browser,
> it seems like both the View and LottoHandler need to change!
>
> is it also the View's responsibility to decide whether to pass error information as strings or as images? 😄

i think `"Enter the numbers for manual tickets."` is view's responsibility!

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,37 +1,32 @@
...
 
     init {
-        require(purchaseAmount in MIN..MAX)
-        change = purchaseAmount % TICKET_PRICE
-        ticketCount = (purchaseAmount - change) / TICKET_PRICE
+        change = ticketCounter.purchaseAmount.amount % TICKET_PRICE
         generateTickets()
     }
 
     fun showChange() = change
```
Hyeon9mak: u can replace the `showChange()` function!

```kotlin
var change = 0
  private set
```

or

```kotlin
private var change = 0
val showChange = change
```

src/main/kotlin/lotto/PurchaseAmount.kt
```diff
@@ -0,0 +1,13 @@
...
+@JvmInline
+value class PurchaseAmount(val amount: Int) {
+    init {
+        require(amount in MIN..MAX) { "Purchase amount must be between $MIN and $MAX" }
+    }
+
+    companion object {
+        private const val MIN = 1_000
+        private const val MAX = 20_000
+    }
```
Hyeon9mak, line 12: 💯

src/test/kotlin/lotto/PurchaseAmountTest.kt
```diff
@@ -0,0 +1,20 @@
...
+    @ValueSource(ints = [-1000, 0, 999, 20_001, 21_000])
+    fun `Purchase amount should be minimum 1_000 and (maximum 20_000 KRW)`(number: Int) {
+        assertThrows<IllegalArgumentException> { PurchaseAmount(number) }
+    }
+
+    @ParameterizedTest
+    @ValueSource(ints = [1_001, 19_000])
+    fun `Purchase amount should be in range of 1_000 and (maximum 20_000 KRW)`(number: Int) {
+        assertEquals(PurchaseAmount(number).amount, number)
+    }
```
Hyeon9mak: 🤔 ... maybe `1_000`, `20_000` ?

src/test/kotlin/lotto/EnteredTicketCountTest.kt
```diff
@@ -0,0 +1,17 @@
...
+class EnteredTicketCountTest {
+    @Test
+    fun `Manual ticket count cannot be less than 0`() {
+        assertThrows<IllegalArgumentException> { EnteredTicketCount(-1) }
+    }
+
+    @Test
+    fun `Manual ticket count must be positive`() {
+        assertEquals(EnteredTicketCount(4).count, 4)
+    }
```
Hyeon9mak: It has to be positive, does it succeed at `0`?
It would be more trust to include a test for these cases.

src/main/kotlin/lotto/TicketCounter.kt
```diff
@@ -0,0 +1,18 @@
...
+) {
+    var generatedTicketCount = 0
+
+    init {
+        var purchasableTickets = purchaseAmount.amount / LottoMachine.TICKET_PRICE
+        require(enteredTicketCount.count <= purchasableTickets) { COUNT_ERROR_MESSAGE }
+        generatedTicketCount = purchasableTickets - enteredTicketCount.count
+    }
+
+    companion object {
```
Hyeon9mak: The reason the current code feels visually complex is...
`LottoMachine` and `TicketCounter` have a dependency on each other.

This is not only visually complex, will also issues during maintenance.
When changes occur on one side, code changes on the other side will inevitably follow.

Therefore, it seems necessary to refactor so that the dependency flows in only one direction.

I suspect - that TicketCounter appears to be a DTO(Data Transfer Object) that passes user input values.
So how about trying to make changes such as minimizing the logic and transferring to the `LottoMachine`?

src/main/kotlin/lotto/TicketCounter.kt
```diff
@@ -0,0 +1,18 @@
...
+
+data class TicketCounter(
+    val purchaseAmount: PurchaseAmount,
+    val enteredTicketCount: EnteredTicketCount,
+) {
+    var generatedTicketCount = 0
+
+    init {
+        var purchasableTickets = purchaseAmount.amount / LottoMachine.TICKET_PRICE
+        require(enteredTicketCount.count <= purchasableTickets) { COUNT_ERROR_MESSAGE }
```
Hyeon9mak:
```suggestion
    init {
        val purchasableTickets = purchaseAmount.amount / LottoMachine.TICKET_PRICE
        require(enteredTicketCount.count <= purchasableTickets) { COUNT_ERROR_MESSAGE }
```

😉

Hyeon9mak (changes requested):
Hi Shahbaz! Nice work to step2. 👍 
Below are the answers to your questions.

### How can Improve my test cases

I think it would be best to conduct tests with **clear success cases, failure cases, and boundary values**.
I believe the best approach is to write tests for points that fellow developers might wonder about when reading the code, thinking 'What would happen in this case?

### Is there any class that can be wrapped or created into another class?

Just like with test improvements, the decision to use a wrapping class should also be based on needs - such as wanting to consolidate business requirements for easy overview or to manage them more conveniently during maintenance. 

Otherwise, there's no need to unnecessarily increase complexity by wrapping.

This is also why developers need not only technical skills but also a deep understanding of business.

### When do I know when I have done too much nesting of class?

The example in the question seems to be more about issues caused by consecutive getter calls rather than too much nesting.
This structure breaks encapsulation, which is highly valued in OOP, where changes to internal objects affect all parent objects.

To improve this issue, instead of calling getters step by step, it would be better to create a separate function that can return the desired value all at once.
For example, like this:

```kotlin
kotlinmachine.ticketPurchaseAmount()
```

### is the use of data and value class to the standards?

Let's think about money as an example for value class.
Most people would consider two $100 bills as the same $100,
but an Finance domain engineer would think of them differently based on each bill's serial number.

In common domains, you could make value class with $100,
but in the Finance domain, you shouldn't use value class in this case.

Let's think about iPhone as an example for data class.
Most people would consider `256GB/space gray/iPhone15` and `256GB/space gray/iPhone15` as the same iPhone,
but an Apple engineer would think of them differently based on each device's serial number.

In common domains, you could make the data class with 3 properties,
but within Apple engineer shouldn't use data class in this case.

You can distinguish use cases based on the concerns of the project.

### minor problem

check this 😉 

image

Does this provide a sufficient answer? 😄 
Feel free to ask any, more questions you have!

src/main/kotlin/lotto/TicketCounter.kt
MehrShahbaz: So I moved the Logic into LottoMachine as it was making it tightly coupled too get the values and making the code more complex let me know If there is something I could have done better

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,37 +1,41 @@
 package lotto
 
 class LottoMachine(
-    val purchaseAmount: Int,
+    val purchaseAmount: PurchaseAmount,
+    val enteredTicketCount: EnteredTicketCount,
```
Hyeon9mak, line 5: The dependency on `LottoMachine` is now isolated to just `LottoHandler`!
It looks like you've minimized the chances of `LottoMachine` being affected by changes in other classes going forward.

The reason this change is important is, because LottoMachine contains a lot of core logic.
Classes with core logic tend to go through frequent and complex changes, and if they're not kept independent, it can make maintenance really difficult.

Great work! 👍

src/main/kotlin/lotto/Lotto.kt
Hyeon9mak: I think `Lotto` can use the `Set<LottoNumber>`.
It's always need distinct numbers.
How do u think? 😉

src/main/kotlin/lotto/LottoHandler.kt
```diff
@@ -1,11 +1,34 @@
...
+            val machine =
+                LottoMachine(
+                    purchaseAmount,
+                    enteredTicketCount,
+                )
+            machine.tickets.addAll(
+                readManualTickets(
+                    machine.manualTicketCount(),
+                ),
+            )
```
Hyeon9mak: > Don’t use any getters/setters/properties. Send message to Object.

How about ask the message to `LottoMachine`?

```suggestion
            val machine = LottoMachine(
                purchaseAmount,
                enteredTicketCount,
            )
            val manualTickets = readManualTickets(machine.manualTicketCount())
            machine.appendManualTickets(manualTickets)
```

src/main/kotlin/lotto/WinningNumbers.kt
```diff
@@ -5,6 +5,11 @@ data class WinningNumbers(
     val bonusNumber: LottoNumber,
 ) {
     init {
-        require(!winningNumbers.numbers.contains(bonusNumber))
+        require(!winningNumbers.numbers.contains(bonusNumber)) { ERROR_MESSAGE }
+    }
```
Hyeon9mak: How about make negative ask function?
It can be remove `!` operator. The `!` operator can be easy to miss.

```kotlin
require(winningNumbers.doesNotContain(bounusNumber)) { ... }
```

Hyeon9mak (changes requested):
Hi Shahbaz!

I checked your changes, and I can see you worked hard to make `LottoMachine` more readable.
It wasn’t easy request, but I’m really happy with how you handled it.

Some parts of the code may not look very simple right now,
but I’m focusing more on writing code that is easy to maintain in the future, not just easy to read now.

I think if you keep trying this kind of coding and improve it later,
you’ll build your own style, and it will become easier to write maintainable code.

There are not many things to fix now,
but I’ll leave 'Request changes' just in case you have questions. 😉 

If you’re okay with everything, please re-request the review.
I’ll approve it right away! 😄

src/main/kotlin/lotto/Lotto.kt
MehrShahbaz: Before I was using Lotto(val numbers: Set<Int>) and using the set was failing for the test 
Lotto(setOf(1, 2, 3, 4, 5, 6, 6)) as it removes an extra 6, which should be a error and when using LottoNumber will also skip this test case so that is why I stayed with list
please share your thoughts on this

Added a small test to check and it should throw and error for length, and unique
Like that is what I am trying the user to do to enter unique numbers
Screenshot 2025-07-09 at 6 11 06 PM

Hyeon9mak: I’m a bit confused by the test example. 
Don’t users usually enter `LottoNumbers` and `BonusNumber` separately?
I’m not sure if there’s a case where all 7 numbers would be passed at once. 
Could you share what concern you were thinking of here?

bellow my example test code.

```kotlin
@Test
fun `Lotto numbers must be 6`() {
    assertThrows<IllegalArgumentException> { Lotto(numbers = mapToSetLottoNumbers(1, 2, 3, 4, 5)) }
    assertDoesNotThrow { Lotto(numbers = mapToSetLottoNumbers(1, 2, 3, 4, 5, 6)) }
    assertThrows<IllegalArgumentException> { Lotto(numbers = mapToSetLottoNumbers(1, 2, 3, 4, 5, 6, 7)) }
}

private fun mapToSetLottoNumbers(vararg numbers: Int): Set<LottoNumber> = numbers.map(LottoNumber::from).toSet()
```

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -1,13 +1,11 @@
...
-        require(numbers.distinct().size == LOTTO_SIZE)
-    }
+class Lotto(numbers: List<LottoNumber>) {
+    val numbers = numbers.toSet()
 
-    fun sortedList(): List<Int> {
-        return numbers.map { it -> it.value }.sorted()
+    init {
+        require(numbers.size == LOTTO_SIZE) { "List should have exactly 6 numbers" }
+        require(this.numbers.size == numbers.size) { "All the numbers in the list must be unique" }
```
Hyeon9mak, line 8: 😉 👍

Hyeon9mak (approved):
Cool!

I misunderstood your intention at first, so it took me a bit longer to review.
Sorry about that. 😅 
But thank you for patiently explaining everything until the end. 👍 

All the requirements for the Lotto mission have been met, and your code looks very clean and readable.
So, I’m happy to approve it!

Great work throughout this mission!
I hope you enjoy the Blackjack mission as well!

PR #19, opened 2025-07-07
[Lotto - Step2] Submit AnnLvu's mission

AnnLvu:
Hi Dani!
This is my pull request for Step 2.
I've added manual ticket input.
I had a big question about whether tickets entered manually from the keyboard could be the same as those generated automatically. In the end, I decided that yes, this could happen. Also, two identical tickets could be entered manually from the keyboard.
Perhaps you have a different opinion on this. What do you think?
I've also fixed your previous comments.
I'm looking forward to new suggestions on what I can improve.
Also, when I was refactoring fun start(), I learned about the Triple class. That helped me a lot!


src/main/kotlin/lotto/controller/LottoMachine.kt
```diff
@@ -1,20 +1,28 @@
 package lotto.controller
 
+import lotto.model.Const
 import lotto.model.Tickets
 import lotto.model.WinningTicket
 import lotto.view.InputView
 import lotto.view.OutputView
 
 object LottoMachine {
     fun start() {
```
da-nyee, line 10: image

Everything works very well 👍

src/main/kotlin/lotto/controller/LottoMachine.kt
```diff
@@ -1,20 +1,28 @@
...
+        OutputView.displayTickets(tickets, manualTicketCount)
         val winningNumbers = InputView.inputWinningNumbers()
         val bonusNumber = InputView.inputBonusNumber(winningNumbers)
         val winningTicket = WinningTicket(winningNumbers, bonusNumber)
         val winStats = tickets.calculateStats(winningTicket)
-
         OutputView.displayResults(winStats, tickets.calculateReturnRate(amount, winStats))
     }
+
+    private fun purchaseTickets(): Triple<Tickets, Int, Int> {
```
da-nyee: What about wrapping the response rather than using `Triple`?

src/main/kotlin/lotto/controller/LottoMachine.kt
```diff
@@ -1,20 +1,28 @@
...
-        val tickets = Tickets.generate(amount)
-        OutputView.displayTickets(tickets)
+        val (tickets, manualTicketCount, amount) = purchaseTickets()
+        OutputView.displayTickets(tickets, manualTicketCount)
         val winningNumbers = InputView.inputWinningNumbers()
         val bonusNumber = InputView.inputBonusNumber(winningNumbers)
         val winningTicket = WinningTicket(winningNumbers, bonusNumber)
         val winStats = tickets.calculateStats(winningTicket)
-
         OutputView.displayResults(winStats, tickets.calculateReturnRate(amount, winStats))
```
da-nyee: The code looks clean and readable 👍

src/main/kotlin/lotto/model/Tickets.kt
```diff
@@ -2,17 +2,23 @@ package lotto.model
 
 class Tickets(private val ticketList: List<Ticket>) {
     companion object {
-        fun generate(amount: Int): Tickets {
-            val ticketCount = amount / Const.PRICE
-            val list =
-                List(ticketCount) {
+        fun generate(
+            amount: Int,
```
da-nyee, line 6: How about wrapping this as well?

src/main/kotlin/lotto/model/Tickets.kt
```diff
@@ -2,17 +2,23 @@ package lotto.model
...
+            manualTickets: List<Numbers> = emptyList(),
```
da-nyee:
```suggestion
            manualTicketNumbers: List<Numbers> = emptyList(),
```

I think it just represents numbers, so I'd recommend renaming the variable.

src/main/kotlin/lotto/model/Tickets.kt
```diff
@@ -28,18 +34,24 @@ class Tickets(private val ticketList: List<Ticket>) {
...
         }
         return winStats
     }
 
+    private fun determineRank(
+        ticket: Ticket,
+        winningTicket: WinningTicket,
+    ): Rank {
+        val matchCount = ticket.numbers.countMatches(winningTicket.winningNumbers)
+        return if (matchCount == 5 && ticket.numbers.contains(winningTicket.bonusNumber)) {
```
da-nyee: How about extracting this into a method and giving it a name?

src/test/kotlin/lotto/InputManualParserTest.kt
```diff
@@ -0,0 +1,64 @@
+package lotto
+
+import lotto.view.InputManualParser
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+
+class InputManualParserTest {
```
da-nyee, line 9: You added a new test! Great work 👍

da-nyee: Consider testing `Tickets` too.

da-nyee (changes requested):
Hi Ann! 🙌
You did a great job — I just left a few comments.
Hope they’re helpful! Let me know if you have any other questions.

---

> I had a big question about whether tickets entered manually from the keyboard could be the same as those generated automatically. In the end, I decided that yes, this could happen. Also, two identical tickets could be entered manually from the keyboard.

Yes, I think it could happen.
However, since there's no rule that prevents generating the same numbers, I believe this situation is acceptable.
If you'd like to make it more strict, you could consider throwing an exception when this occurs.

src/main/kotlin/lotto/controller/LottoMachine.kt
AnnLvu: Yes!
Now, I'm create data class PurchaseResult (val tickets: Tickets,
    val manualTicketCount: Int,
    val amount: Int). 
    And I can use it "purchaseTickets(): PurchaseResult"
    It really looks much more readable and simpler!

src/main/kotlin/lotto/controller/LottoMachine.kt
AnnLvu: Thank you!

src/main/kotlin/lotto/model/Tickets.kt
AnnLvu: Done!

src/main/kotlin/lotto/model/Tickets.kt
AnnLvu: Create new method 'isSecondRank'

src/test/kotlin/lotto/InputManualParserTest.kt
AnnLvu: Added even more tests for Tickets

src/main/kotlin/lotto/model/Tickets.kt
da-nyee: I don't think there is a class named `Amount` or anything like that 🥲

da-nyee (approved):
Hi Ann! 🙌
Thanks for considering the reviews I left 🙇‍♀️

I think it's good to merge now, since you've built the Lotto system perfectly.
Hopefully, the feedback was insightful — cheers to the next mission! 🤸‍♀️

PR #20, opened 2025-07-07
[Lotto - Step 2] Submit arekiu's mission 

arekiu:
Good morning/afternoon/night for the reviewer!

#Lotto machine simulator implementation

## Implement Lotto machine adding new features for manual ticket creation
- Ask the user for amount of custom tickets to be generated
- If the amount is 0, all the tickets will be randomly generated
- Ask the user for the numbers to be included in each ticket
- Generate the rest of tickets (in case there is) with numbers randomly generated
- Display all the tickets

## About implementation

- All the program was refactored to include the LottoNumber class
- Before, LottoManager was created to take responsabilities from LottoMachine because that class was full of methods but now is cleaner. Should I merge LottoManager with LottoMachine, or is it good to keep the classes?
- I would like to know if Validator is being used in the right way (it´s only called by InputView to check and clean the userInput)
- There are a lot of variables in LottoController. Does it look clean and correct? Or should I use more states in classes?

Thank you so much for your time reviewing my code!


src/main/kotlin/lotto/controller/LottoController.kt
```diff
@@ -1,39 +1,49 @@
...
-import lotto.model.TicketsSeller
+import lotto.model.*
 import lotto.view.InputView
 import lotto.view.ResultView
 
 class LottoController {
     private val inputView = InputView()
     private val resultView = ResultView()
     private val ticketSeller = TicketsSeller()
+    private val ticketGenerator = TicketGenerator()
```
Wordbe: Try moving the internal variables declared inside `LottoController` into the primary constructor and injecting them from the outside.
This will make your code more flexible and easier to modify.

This approach is called dependency injection, and it’s definitely worth looking into! 😊

src/main/kotlin/lotto/controller/LottoController.kt
```diff
@@ -1,39 +1,49 @@
...
+        val prizeNumbers = createPrizeNumbers()
+        val dataLottoMachine = LottoMachine(allTickets, prizeNumbers)
+        handleResults(dataLottoMachine, userAmount)
     }
 
-    fun handleUserInput(): LottoMachine {
-        userAmount = inputView.takePurchaseInput()
-        val tickets = ticketSeller.createRandomTickets(userAmount)
-        resultView.printLottos(tickets)
+    fun buyTickets(userAmount: Int): List<LottoTicket> {
```
Wordbe: Please mark methods that are only used internally as private.
Also, whenever you create a private method, consider whether it might be better placed in another class.

This can make your code more testable and lead to a cleaner, more well-structured design! 😊

src/main/kotlin/lotto/model/LottoManager.kt
```diff
@@ -11,18 +11,35 @@ class LottoManager {
...
         tickets.forEach { ticket ->
-            ticket.compareTicket(prizeNumbers.winningNumbers, prizeNumbers.bonusNumber)
+            compareTicket(ticket, prizeNumbers)
             val chosenRank = Rank.getRank(ticket.numberOfHits, ticket.hasBonus)
             prizeCounter[chosenRank] = prizeCounter.getValue(chosenRank) + 1
         }
         return prizeCounter
     }
 
+    fun compareTicket(lottoTicket: LottoTicket, prizeNumbers: PrizeNumbers) {
```
Wordbe: LottoManager is handling quite a lot of logic related to List<LottoTicket> lottoTickets.
How about trying an "inline collection" (also known as a first-class collection) for this?

By wrapping the list in its own class:

You can encapsulate related behaviors, Keep your domain logic clean, And improve readability and maintainability overall. 

It’s a great pattern to explore! 😊

src/main/kotlin/lotto/model/LottoNumber.kt
```diff
@@ -0,0 +1,18 @@
+package lotto.model
+
+class LottoNumber private constructor(private val value: Int) {
+
+    override fun toString(): String = value.toString()
+
+    companion object {
+        private const val MINIMUM_NUMBER = 1
+        private const val MAXIMUM_NUMBER = 45
+        private val NUMBERS: Map<Int, LottoNumber> =
```
Wordbe, line 9: You’ve thoughtfully applied a strategy to prevent uncontrolled instance creation — that’s really well done! 😊

src/main/kotlin/lotto/model/LottoTicket.kt
```diff
@@ -0,0 +1,17 @@
+package lotto.model
+
+open class LottoTicket(
```
Wordbe: Why did you use `open`? 🙂

src/main/kotlin/lotto/model/TicketGenerator.kt
```diff
@@ -2,6 +2,9 @@ package lotto.model
 
 class TicketGenerator() {
 
-    fun createLottoTicket(lottoNumbers: List<Int>) = Lotto(lottoNumbers)
+    fun generateLottoNumbers(numbers: List<Int>) = numbers.map { number -> LottoNumber.from(number) }
```
Wordbe: If it’s not being used externally, marking it as private would be more appropriate. 😊

src/main/kotlin/lotto/model/Validator.kt
```diff
@@ -1,48 +1,32 @@
...
-        }
-        if (userInput.count() != userInput.toSet().count()) {
-            throw IllegalArgumentException("Numbers should be unique")
-        }
-        userInput.forEach { item ->
-            if (item.toInt() !in 1..45) {
-                throw IllegalArgumentException("The winning numbers should be between 1 and 45")
-            }
-        }
+    fun validateTicketNumbers(userInput: List<String>) {
```
Wordbe: If it’s only handling input validation, then naming it `InputValidator` would be perfectly appropriate. 😊

Wordbe (changes requested):
Hello!
You’ve done a great job reflecting the new requirements in your code, and I can clearly see that the structure has improved compared to before.

Please take a look at the review, and feel free to apply any suggestions you find helpful.
If you have a different perspective on any of the comments, don’t hesitate to share it!

Great work thank you! 😊

---

Q1 - answer:
If LottoMachine and LottoManager have clearly divided responsibilities, it would be a good idea to define clear boundaries and continue separating their roles based on that criteria. 😊

---

Please check the comments for the rest of my questions — I’ve left them there for you! 😊

src/main/kotlin/lotto/controller/LottoController.kt
arekiu: Done! I´m slowly understanding how the classes should be independent between them and now I see it doesn´t have much sense to create the instances in the inside. Adding dependency injection makes it more flexible for testing for example.

src/main/kotlin/lotto/controller/LottoController.kt
arekiu: Methods changed to private, next time I will put more attention!
However,  I didn´t find a better place to put these methods. My intention is that LottoController is the intermediary between user and LottoMachine. So it´s in charge to generate the tickets (in buyTickets()) and get the winning numbers (in createPrizeNumbers()) to inject in LottoMachine in order to process them.

Please, let me know if you think it´s a better idea to create a separate class.

src/main/kotlin/lotto/model/LottoNumber.kt
arekiu: I can´t take credit for that, because it was part of the lecture feedback. But, for sure I´ll implement it again in next projects!

src/main/kotlin/lotto/model/TicketGenerator.kt
arekiu: Before it was public because it was called to convert the winning numbers given by the user into a list of LottoNumbers (see the answer of why I used " open " for the class). But now, I changed the structure and the winning numbers will be directly converted into a LottoTicket

src/main/kotlin/lotto/model/LottoTicket.kt
arekiu: The class PrizeNumber was inheritating from it, having a LottoTicket (with winning numbers) plus the Bonus number.
But I realized it didn´t make sense and now I´m using "composition" to achieve the same. So LottoTicket is not open anymore

`class PrizeNumbers(
    val winningNumbersTicket: LottoTicket,
    val bonusNumber: LottoNumber
)`

src/main/kotlin/lotto/model/LottoManager.kt
arekiu: Great suggestion! I created LottoTickets class to wrap all the LottoTicket. I also move the methods from LottoMachine, like evaluateAll() to LottoTickets and compareTicket() to LottoTicket and now is much more clear.
In the other hand, LottoMachine and LottoControler had not much work to do, so I decided to replace them by the class LottoEvaluator that will be in charge of processing the result of all the evaluated tickets.

arekiu:
Thank you again for your detailed review! Let me know if it´s necessary more refactoring.
Currently I´m starting the second mission BlackJack with better organization and planning to have a cleaner architecture than in Lotto.
Have a nice day!

src/main/kotlin/lotto/Application.kt
```diff
@@ -1,9 +1,23 @@
...
 
 fun main() {
-    val lottoController = LottoController()
+    val inputView = InputView()
+    val resultView = ResultView()
+    val ticketSeller = TicketsSeller()
+    val ticketGenerator = TicketGenerator()
 
-    lottoController.runMachine()
+    val lottoController = LottoController(
```
Wordbe: Looks great 👍

src/main/kotlin/lotto/controller/LottoController.kt
```diff
@@ -1,39 +1,51 @@
...
-    fun handleUserInput(): LottoMachine {
-        userAmount = inputView.takePurchaseInput()
-        val tickets = ticketSeller.createRandomTickets(userAmount)
-        resultView.printLottos(tickets)
+    private fun buyTickets(userAmount: Int): List<LottoTicket> {
+        var numberOfTickets = ticketSeller.calculateTickets(userAmount)
+        val numberOfManualTickets = inputView.takeNumberOfCustomTickets(numberOfTickets)
+        val customTicketNumbers = inputView.takeCustomLottoNumbers(numberOfManualTickets)
+        val customTickets = ticketSeller.createCustomTickets(customTicketNumbers)
+        numberOfTickets -= numberOfManualTickets
```
Wordbe: It would be better to define the new variable using val instead of var.
This allows you to give it a more descriptive name and ensures immutability, which reduces potential risks in your code. 😊

src/main/kotlin/lotto/controller/LottoController.kt
Wordbe: That sounds great. I agree with the current placement as well!

src/main/kotlin/lotto/controller/LottoController.kt
```diff
@@ -1,39 +1,51 @@
...
     }
 
-    fun handleResults(lottoMachine : LottoMachine) {
-        val ticketsResult = lottoMachine.evaluateTicketsResult()
-        resultView.printResult(ticketsResult)
-
-        val returnRate = lottoMachine.evaluateReturnRate(ticketsResult, userAmount)
-        resultView.printReturnRate(returnRate)
+    private fun handleResults(allTickets: LottoTickets, prizeNumbers: PrizeNumbers, userAmount: Int) {
+        val lottoEvaluator = LottoEvaluator()
```
Wordbe: I think it would be good if LottoEvaluator also used dependency injection.
That way, it becomes more flexible and easier to test! 😊

src/main/kotlin/lotto/controller/LottoController.kt
```diff
@@ -1,39 +1,51 @@
...
+        val ticketsResult = allTickets.evaluateAll(prizeNumbers)
```
Wordbe: How about handling this inside LottoEvaluator as well?
It could help keep related logic encapsulated and improve cohesion! 😊

src/main/kotlin/lotto/model/LottoManager.kt
Wordbe: Great. Thank you!

src/main/kotlin/lotto/model/LottoTicket.kt
Wordbe: Using composition instead of inheritance - great improvement!

Wordbe (changes requested):
Hello Alex!

Thank you for quickly reflecting the review feedback.
I can see that the changes you’ve made are gradually shaping the code into a more readable and maintainable structure!

I’ve left some comments with suggestions worth considering, so please take a look when you have a chance.

Thanks again! 😊

src/main/kotlin/lotto/controller/LottoController.kt
arekiu: Right! I did a modification and now is clear that there are 3 different values:

val numberTotalOfTickets
val numberOfManualTickets
val numberOfRandomTickets

src/main/kotlin/lotto/controller/LottoController.kt
arekiu: The only reason why I didn´t apply this before is because in that way, I need to mock val results: Map<Rank, Int> before, to make tests using the LottoEvaluator methods. 
Before the injection dependency, the methods calculateReturnRate() and calculateTotalPrize() were open , so I could call them for small tests.
Now they are private and called by the method fun evaluateReturnRate().

Again, I don´t feel super happy about my architecture and I´m doing my best to improve this during BlackJack mission.

src/main/kotlin/lotto/controller/LottoController.kt
arekiu: Actually, I realized that method is unnecessary. Now LottoEvaluator takes care of handling the results.

arekiu:
Thank you again for the review, it helps me a lot to realize what things were not a good choice, so I can improve them.
Please, let me know if there is something else I can do better!

src/main/kotlin/lotto/controller/LottoController.kt
```diff
@@ -1,39 +1,50 @@
...
 import lotto.view.ResultView
 
-class LottoController {
-    private val inputView = InputView()
-    private val resultView = ResultView()
-    private val ticketSeller = TicketsSeller()
+class LottoController(
+    private val inputView: InputView,
+    private val resultView: ResultView,
+    private val ticketSeller: TicketsSeller,
```
Wordbe: You might want to align the variable name with its type - it looks like an “s” is missing! 😊

src/main/kotlin/lotto/controller/LottoController.kt
```diff
@@ -1,39 +1,50 @@
...
+    private val ticketGenerator: TicketGenerator
```
Wordbe: Since TicketGenerator is ultimately responsible for creating LottoTickets, it seems like LottoTicket itself could take on that role. 😊

src/main/kotlin/lotto/model/LottoTickets.kt
```diff
@@ -0,0 +1,23 @@
+package lotto.model
+
+class LottoTickets(val tickets: List<LottoTicket>) {
+
+    private fun createMap(): MutableMap<Rank, Int> {
```
Wordbe: How about replacing the Map with an object that holds both Rank and Int to make the meaning clearer? 😊

src/main/kotlin/lotto/model/LottoEvaluator.kt
```diff
@@ -0,0 +1,22 @@
+package lotto.model
+
+class LottoEvaluator(
+    val results: Map<Rank, Int>, val userAmount: Int
+) {
+
+    private fun calculateTotalPrize(): Int {
```
Wordbe: Try refactoring it into a functional style! 😊
Kotlin is powerful for this area.

src/main/kotlin/lotto/model/NumberGenerator.kt
```diff
@@ -1,10 +1,8 @@
 package lotto.model
 
-import kotlin.collections.shuffled
-
 class NumberGenerator {
 
-    fun generateNumbers() = (MINIMUM_VALUE..MAXIMUM_VALUE).shuffled().take(6).sorted()
+    fun generateRandomNumbers(): List<Int> = (MINIMUM_VALUE..MAXIMUM_VALUE).shuffled().take(6).sorted()
```
Wordbe: In TicketSeller, instead of directly depending on NumberGenerator, try creating a new interface and depend on that.
Then, make NumberGenerator an implementation of the new interface.

This way, TicketSeller can be tested not only for the count but also for specific lotto numbers.
If you have time, it would be great to give it a try! 😊

Wordbe (changes requested):
Hello Alex!
You're doing a great job, and thank you for actively applying the feedback. 🙏
I’ve added a bit more to the review since I wanted to share even more helpful points.
Please take a look, and if possible, try applying the parts that make sense to you.

Thank you! 😊

src/main/kotlin/lotto/controller/LottoController.kt
arekiu: Right. The name in the class was correct but I may had forget it when naming the val

src/main/kotlin/lotto/controller/LottoController.kt
arekiu: That´s true, LottoGenerator was unnecesary. I had the wrong idea that I should give "Lotto numbers" to LottoTicket in order to create a new Ticket. But now, it receives a list of Int and transform them inside to Lotto numbers without exposing this process. I find it more logical.

src/main/kotlin/lotto/model/LottoTickets.kt
arekiu: I didn´t find a better way to achieve that without using a map. But I created the class LottoResults to store it and I think the functionality now is a little more clear

src/main/kotlin/lotto/model/NumberGenerator.kt
arekiu: It was a little difficult to understand why do I need the interface, but finally I got it. I can create a different class from the inferface and give it to TicketSeller. So, for example, I could test it with fixed numbers.
I will try to consider this for future projects

arekiu:
Good day! I applied the changes and learned some new concepts in the process.
Please, let me know how is the project looking.

Have a nice day!

Wordbe (approved):
Thank you for your thoughtful replies and for applying the feedback!

I hope this Lotto mission has been a meaningful step forward in your growth.
Congratulations on completing it successfully — truly well done! 🎉

I’ll go ahead and merge this PR here.
Looking forward to seeing your great work continue in the next mission as well.
Thank you again, and great job! 😊

PR #22, opened 2025-07-07
[Lotto - Step2] Submit Mariia's mission

mariyakulikova:
Hi!
This is my first approach for Step 2. I've added manual input for Lotto tickets.

While working on this step, I’ve run into a concern and would really appreciate your feedback and suggestions:

I currently have two functions in LottoMachine that essentially do the same thing, they both return a list of numbers for a Lotto ticket.
I feel like this should be improved, maybe by introducing an interface, but I’m not quite sure how to approach it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,8 @@ Update the winning statistics to include the second prize category.
 - [x] get input of amount
 - [x] get the winning numbers
 - [x] get the bonus number
+- [ ] get number of manuel tickets
+- [ ] get numbers for manuel tickets
 
 #### Output result
 - [x] should print the number of purchased tickets
--- a/README.md
+++ b/README.md
@@ -19,7 +19,8 @@ Update the winning statistics to include the second prize category.
 - [x] lotto ticket each number should be between 1 and 45
 
 - [x] generate 6 unique, random, numbers in range 1 to 45
-- [x] machine issues lotto tickets
+- [x] machine issues lotto tickets with generated numbers
+- [ ] machine issues lotto tickets with manuel provided numbers 
 - [x] figure out winning tickets
 
 #### Input
--- a/README.md
+++ b/README.md
@@ -27,8 +27,8 @@ Update the winning statistics to include the second prize category.
 - [x] get input of amount
 - [x] get the winning numbers
 - [x] get the bonus number
-- [ ] get number of manuel tickets
-- [ ] get numbers for manuel tickets
+- [x] get number of manuel tickets
+- [x] get numbers for manuel tickets
 
 #### Output result
 - [x] should print the number of purchased tickets
--- a/src/main/kotlin/lotto/Application.kt
+++ b/src/main/kotlin/lotto/Application.kt
@@ -5,9 +5,16 @@ import view.ResultView
 
 fun main() {
     val purchaseAmount = Money(InputView.getPurchaseAmount())
-    val purchasedTicketNumber = PurchasedTicketsNumber(InputView.getNumberOfManuelTickets(), purchaseAmount)
-    val lottoMachine = LottoMachine(purchasedTicketNumber)
-    val lottoTickets = lottoMachine.issueTickets()
+    val lottoMachine = LottoMachine()
+    val manuelTicketsNumber = InputView.getNumberOfManuelTickets()
+    val totalTicketsNumber = purchaseAmount.value / TICKET_PRICE
+    val lottoTickets =
+        lottoMachine.issueTickets(manuelTicketsNumber) {
+            InputView.getNumbersForTicket()
+        } +
+            lottoMachine.issueTickets(totalTicketsNumber - manuelTicketsNumber) {
+                Generator.randomNumber()
+            }
 
     ResultView.viewTicketsAmount(lottoTickets.size)
     ResultView.viewTickets(lottoTickets)
--- a/src/main/kotlin/lotto/Application.kt
+++ b/src/main/kotlin/lotto/Application.kt
@@ -18,5 +18,5 @@ fun main() {
     val game = LottoGame()
     val gameResults = game.play(lottoTickets, winningNumbers, bonusNumber)
     ResultView.viewWinning(gameResults)
-    ResultView.viewRate((game.countWinningAmount(gameResults) / purchaseAmount.value).toDouble())
+    ResultView.viewRate((game.countWinningAmount(gameResults) / purchaseAmount.getValue()).toDouble())
 }
--- a/src/main/kotlin/lotto/Application.kt
+++ b/src/main/kotlin/lotto/Application.kt
@@ -4,7 +4,7 @@ import view.InputView
 import view.ResultView
 
 fun main() {
-    val amountInput = InputView.getAmountInput()
+    val amountInput = InputView.getPurchaseAmount()
     val purchaseAmount = Money(amountInput)
     val lottoMachine = LottoMachine(purchaseAmount)
     val lottoTickets = lottoMachine.tickets
--- a/src/main/kotlin/lotto/Application.kt
+++ b/src/main/kotlin/lotto/Application.kt
@@ -13,7 +13,7 @@ fun main() {
     ResultView.viewTickets(lottoTickets)
 
     val winningNumbers = Lotto(InputView.getWinningNumbers())
-    val bonusNumber = InputView.getBonusNumber()
+    val bonusNumber = LottoNumber.from(InputView.getBonusNumber())
 
     val game = LottoGame()
     val gameResults = game.play(lottoTickets, winningNumbers, bonusNumber)
--- a/src/main/kotlin/lotto/Application.kt
+++ b/src/main/kotlin/lotto/Application.kt
@@ -4,9 +4,9 @@ import view.InputView
 import view.ResultView
 
 fun main() {
-    val amountInput = InputView.getPurchaseAmount()
-    val purchaseAmount = Money(amountInput)
-    val lottoMachine = LottoMachine(purchaseAmount)
+    val purchaseAmount = Money(InputView.getPurchaseAmount())
+    val purchasedTicketNumber = PurchasedTicketsNumber(InputView.getNumberOfManuelTickets(), purchaseAmount)
+    val lottoMachine = LottoMachine(purchasedTicketNumber)
     val lottoTickets = lottoMachine.tickets
 
     ResultView.viewTicketsAmount(lottoTickets.size)
--- a/src/main/kotlin/lotto/Application.kt
+++ b/src/main/kotlin/lotto/Application.kt
@@ -18,5 +18,5 @@ fun main() {
     val game = LottoGame()
     val gameResults = game.play(lottoTickets, winningNumbers, bonusNumber)
     ResultView.viewWinning(gameResults)
-    ResultView.viewRate((game.countWinningAmount(gameResults) / purchaseAmount.getValue()).toDouble())
+    ResultView.viewRate((game.countWinningAmount(gameResults).value / purchaseAmount.value).toDouble())
 }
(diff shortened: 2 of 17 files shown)


README.md
```diff
@@ -19,13 +19,16 @@ Update the winning statistics to include the second prize category.
 - [x] lotto ticket each number should be between 1 and 45
 
 - [x] generate 6 unique, random, numbers in range 1 to 45
-- [x] machine issues lotto tickets
+- [x] machine issues lotto tickets with generated numbers
+- [ ] machine issues lotto tickets with manuel provided numbers
```
Wordbe, line 23: Check this please if you already completed!

src/main/kotlin/lotto/Application.kt
```diff
@@ -4,19 +4,19 @@ import view.InputView
...
 
     val winningNumbers = Lotto(InputView.getWinningNumbers())
-    val bonusNumber = InputView.getBonusNumber()
+    val bonusNumber = LottoNumber.from(InputView.getBonusNumber())
 
     val game = LottoGame()
     val gameResults = game.play(lottoTickets, winningNumbers, bonusNumber)
     ResultView.viewWinning(gameResults)
-    ResultView.viewRate((game.countWinningAmount(gameResults) / purchaseAmount.value).toDouble())
+    ResultView.viewRate((game.countWinningAmount(gameResults) / purchaseAmount.getValue()).toDouble())
```
Wordbe: If the value property in Money is marked as private, it’s better not to access it directly.
Instead, consider adding operation methods inside the Money class itself.
Also, it would be great if game.countWinningAmount returned a Money object as well — that would keep the logic encapsulated and consistent! 😊

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -2,25 +2,24 @@ package lotto
 
 private const val NUMBERS_SIZE = 6
 
-class Lotto(val numbers: List<Int>) {
+class Lotto(numbersList: List<Int>) {
```
Wordbe: If you let the primary constructor take a List<LottoNumber> and have the secondary constructor accept a List<Int>, you can avoid creating unnecessary internal variables.
It’ll keep the class cleaner and more focused! 😊

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,15 +1,31 @@
...
 
-        return (1..ticketCount).map {
-            Lotto(generateRandomNumber())
-        }
+    private fun generateRandomNumber(): List<Int> {
+        return (LottoNumber.MINIMUM_NUMBER..LottoNumber.MAXIMUM_NUMBER).shuffled().take(6)
     }
 
-    private fun generateRandomNumber() = (1..45).shuffled().take(6).sorted()
+    private fun manuelInputNumbers(): List<Int> = InputView.getNumbersForTicket()
```
Wordbe: It would be ideal to structure your code so that the view layer doesn’t leak into the domain layer.
Previously, the Application layer connected both view and domain, but consider refactoring so that these boundaries remain clearly separated.

Try designing it so that:

The domain contains only pure business logic, The view handles only I/O or presentation, And the application coordinates them without breaking separation. 

Maintaining this separation will make your codebase more modular, testable, and maintainable. 😊

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,15 +1,31 @@
 package lotto
 
-class LottoMachine(val money: Money) {
+import view.InputView
+
+class LottoMachine(val ticketsNumber: PurchasedTicketsNumber) {
```
Wordbe: The tickets variable is currently stored as an internal state — but is that really necessary?
Could it instead be returned directly from a method and handled externally?

If there’s no strong reason to retain it as state, returning it as a result would make the class more predictable and easier to manage, reducing side effects.
It’s worth trying! 😊

src/main/kotlin/lotto/LottoNumber.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    override fun toString(): String {
+        return value.toString()
+    }
+
+    companion object {
+        const val MINIMUM_NUMBER = 1
+        const val MAXIMUM_NUMBER = 45
+        private const val ERROR_MSG = "Value is not in range from $MINIMUM_NUMBER to $MAXIMUM_NUMBER"
+        private val NUMBERS: Map<Int, LottoNumber> =
```
Wordbe, line 16: You’ve pre-created the objects to prevent uncontrolled instantiation — that’s a great approach! 😊

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,15 +1,31 @@
...
+            },
+        )
+
+        return tickets.toList()
+    }
 
-        return (1..ticketCount).map {
-            Lotto(generateRandomNumber())
-        }
+    private fun generateRandomNumber(): List<Int> {
```
Wordbe: If you extract the uncontrollable part — like random number generation — to the outside, your code becomes testable!
That means in LottoMachineTest, you won’t just be testing Lotto numbers in isolation, but also the actual outcomes more directly.

Shall we give it a try? 😊

src/main/kotlin/lotto/PurchasedTicketsNumber.kt
```diff
@@ -0,0 +1,19 @@
+package lotto
+
+class PurchasedTicketsNumber(private val manuelTicketsNumber: Int, purchasedAmount: Money) {
+    private val totalTickets = purchasedAmount.getValue() / TICKET_PRICE
```
Wordbe: Let’s avoid creating internal variables — instead, have the method perform the calculation and return the result directly! 😊

Wordbe (changes requested):
Hello!

It’s great to see that you’ve not only implemented the new requirements but also improved the structure compared to before.

Please take a look at the review comments
 feel free to apply what makes sense to you, and of course, different perspectives are always welcome!

Thank you! 😊

src/main/kotlin/lotto/LottoNumber.kt
mariyakulikova: Thank you! 😊
Unfortunately, it’s not my idea — this code is from the lecture materials.

src/main/kotlin/lotto/Application.kt
mariyakulikova: I was thinking of making the value field in the `Money` class private to avoid possible modifications from outside. However, in the getter, I’m apparently returning not a copy of the value, but the value itself, since it’s a primitive type.

I’ve changed the return type of `game.countWinningAmount` to `Money`. Thanks for the idea!

src/main/kotlin/lotto/LottoMachine.kt
mariyakulikova: I agree storing tickets in `LottoMachine` is not necessary. I've refactored `LottoMachine`. Now it returns list of `Lotto`.

src/main/kotlin/lotto/LottoMachine.kt
mariyakulikova: At first sight it seems undoable. But I gave it a try, after I extracted `randomNumberGenerator` the whole picture became visible. Thank you for encouraged me!

src/main/kotlin/lotto/PurchasedTicketsNumber.kt
mariyakulikova: During refactor I removed this class, because it became unnecessary. Thank you!

src/main/kotlin/lotto/LottoMachine.kt
mariyakulikova: If I pass `InputView.getNumbersForTicket` like lambda expression to LottoMachine, will it be also leak into the domain layer?

README.md
Wordbe, line 23: Check this one please🙂

src/main/kotlin/lotto/Application.kt
```diff
@@ -4,19 +4,26 @@ import view.InputView
...
 
 fun main() {
-    val amountInput = InputView.getAmountInput()
-    val purchaseAmount = Money(amountInput)
-    val lottoMachine = LottoMachine(purchaseAmount)
-    val lottoTickets = lottoMachine.tickets
+    val purchaseAmount = Money(InputView.getPurchaseAmount())
+    val lottoMachine = LottoMachine()
+    val manuelTicketsNumber = InputView.getNumberOfManuelTickets()
+    val totalTicketsNumber = purchaseAmount.value / TICKET_PRICE
```
Wordbe, line 10: Instead of accessing purchaseAmount.value directly, try creating and using a divide method inside the Money class. 🙂

src/main/kotlin/lotto/Application.kt
```diff
@@ -4,19 +4,26 @@ import view.InputView
...
-    val winningNumbers = Lotto(InputView.getWinningNumbers())
-    val bonusNumber = InputView.getBonusNumber()
+    val winningNumbers = Lotto(InputView.getWinningNumbers().map { LottoNumber.from(it) })
+    val bonusNumber = LottoNumber.from(InputView.getBonusNumber())
 
     val game = LottoGame()
     val gameResults = game.play(lottoTickets, winningNumbers, bonusNumber)
     ResultView.viewWinning(gameResults)
-    ResultView.viewRate((game.countWinningAmount(gameResults) / purchaseAmount.value).toDouble())
+    ResultView.viewRate((game.countWinningAmount(gameResults).value / purchaseAmount.value).toDouble())
```
Wordbe, line 28: Same here!

src/main/kotlin/lotto/Generator.kt
```diff
@@ -0,0 +1,5 @@
+package lotto
+
+object Generator {
+    fun randomNumber() = (LottoNumber.MINIMUM_NUMBER..LottoNumber.MAXIMUM_NUMBER).shuffled().take(6)
```
Wordbe, line 4: Explicitly specifying the return type of randomNumber makes the code easier to understand. 🙂

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,15 +1,12 @@
...
 
-class LottoMachine(val money: Money) {
-    val tickets: List<Lotto> = issueTickets()
-
-    fun issueTickets(): List<Lotto> {
-        val ticketCount: Int = money.value / TICKET_PRICE
-
-        return (1..ticketCount).map {
-            Lotto(generateRandomNumber())
+class LottoMachine() {
```
Wordbe, line 3: The default constructor is automatically created!
You can omit the parentheses in LottoMachine()!

src/main/kotlin/lotto/LottoMachine.kt
Wordbe: Yes, that's right. In this project, the object that seems to play the role of the controller is LottoGame, so ideally the view should only exist there. 😊

src/main/kotlin/lotto/LottoMachine.kt
Wordbe: Great!

Wordbe (approved):
Hello Mariia!

Thank you for your dedication in completing the Lotto mission to the end.

Including the reviews I provided, I believe many of the improvements that could be made compared to the initial code have been well reflected, so I’ll go ahead and merge it here!

However, I hope you’ll take a moment to review and apply the additional comments I left in the feedback.

Great job on the Lotto mission, and I hope what you’ve studied and learned here will be helpful in your future coding!

Thank you!! 😊

src/main/kotlin/lotto/LottoMachine.kt
mariyakulikova: Thank you!
A model doesn't know anything about view and controller. A controller knows about view and model, so view is called inside a controller and returned value are passed to model. Do I understand right?

PR #24, opened 2025-07-07
[Lotto - Step2] Submit SebasNadu's mission

SebasNadu:
Hi PK 👋

I'm looking forward to reading your feedback!
I'd also love to hear your thoughts on using an immutable PurchaseSession data class, do you find it practical and convenient, or does it feel like overengineering?

I chose to keep the logic for creating manual tickets in the controller, since it's tightly coupled with the user input. In contrast, automatic tickets are still generated in the LottoMachine.
Do you think this split makes sense, or would you handle it differently?

Thanks in advance for your feedback!

## Step 2 Features

* Introduced `Lotto` with support for both `AUTO` and `MANUAL` types.
* Added `PurchaseSession` data class to encapsulate all state related to the purchase process. This replaces previously scattered variables in the controller.
* `PurchaseSession` is immutable and provides convenience methods to update its state via deep copies.
* `LottoMachine` is now solely responsible for generating automatic tickets.
* Refactored the controller to support the new `PurchaseSession`-based flow.
* The controller now uses constructor injection for `LottoMachine` and `PurchaseSession`.
* The program can now support multiple purchase sessions by resetting `purchaseSession` and re-running the controller.
* Added a retry limit to `retryUntilSuccess()` to prevent infinite loops.
* Introduced a maximum purchase amount to prevent unrealistic values.
* Updated tests to support the new `PurchaseSession` system.
* Improved formatting and readability of the `ViewOutput` class.

### TODO

* Add I/O integration tests.


README.md
```diff
@@ -6,6 +6,7 @@
 
 - [x] 6 unique numbers
 - [x] Lotto instance has toString
+- [x] Add type automatic or manual to the class
```
pkeugine, line 9: Nice work updating the README document 👍 

Just by looking at the commit count of this  PR, I was surprised by the amount of commits you've made for this PR 😮 

To be honest, I do feel like it's a bit too much and some commits can be squashed into one.
Creating this many micro commits will make the creator and reviewer (or people who manage them) take more time to follow it.

But hey, [we talked about this](https://github.com/herotechcourse/kotlin-lotto/pull/7#discussion_r2188900712), and we wanted to experiment with various methods right? :)
When working on a new project, are you going to approach in a same way or try out some bigger commits?

Also, it'll be nice to checkout `rebase`, `squash`, and `fixup` features of git.
By using them, we'll be able to manage git commits like pro and even **modify** commits from the past 🪄

src/main/kotlin/lotto/domain/PurchaseSession.kt
```diff
@@ -0,0 +1,57 @@
...
+    val returnRate: Double = 0.0,
+) {
+    val automaticTicketsNumber: Int
+        get() = allTicketsNumber - manualTicketsNumber
+    val allTickets: List<Lotto>
+        get() = manualTickets + automaticTickets
+
+    fun updateAmount(amount: Int): PurchaseSession {
+        val condition = amount >= LottoMachine.PRICE_OF_TICKET && amount % LottoMachine.PRICE_OF_TICKET == 0
+        require(condition && amount <= LottoMachine.MAX_AMOUNT_ACCEPTED) {
```
pkeugine: Only a part of the whole condition was put in the variable, while the last one still exists in `require` block. Also, since there are multiple conditions here, it's kinda hard to understand it right away. Just by dividing them into smaller parts and giving them meaningful name, we'll be able to make it more readable : 
```suggestion
        val isAboveMinimum = amount >= LottoMachine.PRICE_OF_TICKET
        val isDivisible = amount % LottoMachine.PRICE_OF_TICKET == 0
        val isBelowMaximum = amount <= LottoMachine.MAX_AMOUNT_ACCEPTED

        require(isAboveMinimum && isDivisible && isBelowMaximum) {
            throw InvalidAmountException(amount)
        }
```

src/test/kotlin/lotto/domain/PurchaseSessionTest.kt
```diff
@@ -0,0 +1,106 @@
...
+    @Test
+    fun `should update amount and return a new instance from the session`() {
+        val session1 = PurchaseSession()
+        val session2 = session1.updateAmount(45000)
+        assertThat(session2).isNotEqualTo(session1)
+        assertThat(session2.amount).isNotEqualTo(session1.amount)
+    }
+
+    @Test
+    fun `should throw an exception and not update amount due a invalid amount`() {
```
pkeugine:
```suggestion
    fun `should throw an exception and not update amount due to an invalid amount`() {
```

Small, but found a missing word 🤓

src/test/kotlin/lotto/domain/PurchaseSessionTest.kt
```diff
@@ -0,0 +1,106 @@
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+
+class PurchaseSessionTest {
+    @Test
+    fun `should return a deep copy`() {
+        val session1 = PurchaseSession(amount = 1)
+        val session2 = session1.copy(manualTicketsNumber = 10)
+        assertThat(session2).isNotEqualTo(session1)
+        assertThat(session2.amount).isEqualTo(session1.amount)
```
pkeugine:
```suggestion
        assertThat(session2).isNotEqualTo(session1)
            .extracting { it.amount }.isEqualTo(session1.amount)
```

This comment is optional.
I just wanted to show you there's a lot of cool tools that allow us to assert complex situations,
and we can chain them together.
By knowing them, we can tackle harder test scenarios in the future 👍

src/main/kotlin/lotto/domain/Lotto.kt
```diff
@@ -23,8 +23,11 @@ class Lotto(val numbers: Set<LottoNumber>) {
     companion object {
         const val LOTTO_PICK_SIZE = 6
 
-        fun fromInts(numbers: Set<Int>): Lotto {
-            return Lotto(numbers.map(LottoNumber::from).toSet())
+        fun fromInts(
```
pkeugine: This might seem like it's not important, but since `Int` is not an actual word,
IntelliJ sees it as a typo. We do know that there's an `Int` type in Kotlin,
but why don't we use a full name so that it doesn't have the green line underneath?
Yes, we can add it to our IntelliJ dictionary, but teammates might not do it and they'll see the green line.
Since there's an easy replacement, we might as well remove it.

src/main/kotlin/lotto/domain/ResultAnalyzer.kt
```diff
@@ -1,16 +1,15 @@
...
 
-    fun calculateReturnRate(
-        totalAmount: Int,
-        results: List<Rank>,
-    ): Double {
-        val totalPrize = results.sumOf { it.winningMoney }
-        return if (totalPrize == 0) 0.0 else (totalPrize.toDouble() / totalAmount) * 100.0
+    fun calculateReturnRate(session: PurchaseSession): Double {
+        val totalPrize = session.ticketsRank.sumOf { it.winningMoney }
+        return if (totalPrize == 0) 0.0 else totalPrize.toDouble() / session.amount
```
pkeugine, line 13: Nice refactoring 👍 
After the refactor, I think this analyzer utility class can actually be absorbed into `PurchaseSession`, since the only thing it needs is that class. When we want to evaluate tickets and calculate the return rate, we might as well let `PurchaseSession` do it for us rather than using two classes in order to achieve it.

Though , some might say `PurchaseSession` will have too much responsibility, I think that is another topic after we move this into that class.

src/main/kotlin/lotto/exceptions/LottoException.kt
```diff
@@ -37,4 +37,19 @@ sealed class LottoException(msg: String) : IllegalArgumentException("[ERROR]: $m
...
+
+    class InvalidManualTicketsSizeFormatException(input: String) : LottoException(
+        "Invalid manual tickets number format: $input",
+    )
+
+    class InvalidManualTicketsNumberException(amount: Int) : LottoException(
+        "Invalid amount of manual tickets: $amount: should be less or equal to the amount of purchased tickets.",
+    )
+
+    class InvalidManualTicketsException() : LottoException("Invalid manual tickets generations.")
```
pkeugine:
```suggestion
    class InvalidManualTicketsException : LottoException("Invalid manual tickets generations.")
```

Kotlin allows us to keep things simple 😋

src/test/kotlin/lotto/domain/PurchaseSessionTest.kt
```diff
@@ -0,0 +1,106 @@
...
+                            3,
+                            4,
+                            5,
+                            6,
+                        ),
+                    )
+                },
+            )
+        }
+    }
```
pkeugine: Wow, nice job covering a lot of cases with tests 👍

src/test/kotlin/lotto/domain/PurchaseSessionTest.kt
```diff
@@ -0,0 +1,106 @@
+package lotto.domain
+
+import lotto.exceptions.LottoException
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+
+class PurchaseSessionTest {
+    @Test
+    fun `should return a deep copy`() {
```
pkeugine: Although `PurchaseSession` is an immutable object, it isn't actually `deep copying` itself when `copy()` is called. So the test isn't wrong, but the title of the test doesn't seem like what it actually does. Why don't we change the title? :)

Also, this might be a good time to go over the term `deep copy`, `shallow copy`, and `immutability`.

src/main/kotlin/lotto/domain/ResultAnalyzer.kt
pkeugine: **Q. your thoughts on using an immutable PurchaseSession data class**

It's interesting to see this implementation, since it reminds me of `context`s in various kinds of libraries, frameworks, and some other projects.

Keeping these could be helpful, but like how I dropped the hint in above comment, I think it could be separated or simplified. (For example, I don't think it really needs `manualTicketNumber`)

To be honest I do think it's a bit of an over engineering in this level of project because it increases the amount of time for coworker / reviewer to understand the logic. Contexts are often used in cases where we need to pass around the states in between multiple components. I'm more of a 'simple-is-best, make-something-when-you-really-need-it' kind of a guy so this is just my opinion :)

Oh, and I still think it's worth a try, and this kind of implementation does exist in practical usage. 👍

src/main/kotlin/lotto/controller/LottoController.kt
```diff
@@ -3,55 +3,95 @@ package lotto.controller
...
+        if (purchaseSession.manualTicketsNumber == 0) return
+
+        InputView.printGetManualTicketsHeader()
+        purchaseSession =
+            purchaseSession.updateManualTickets(
+                List(purchaseSession.manualTicketsNumber) {
+                    retryUntilSuccess { Lotto.fromInts(InputView.getManualTicket(), LottoType.MANUAL) }
+                },
+            )
+    }
```
pkeugine: **Q. I chose to keep the logic for creating manual tickets in the controller, since it's tightly coupled with the user input. In contrast, automatic tickets are still generated in the LottoMachine.
Do you think this split makes sense, or would you handle it differently?**

I agree with the reason behind putting `updateManualTickets` in the controller while having `updateAutomaticTickets` in `LottoMachine`, and I do think it makes sense.

Like I said in the previous comment, I'm a 'simple-is-best' kind of a guy, and I think in order to maintain the project in a long term, conventions and logic placements should always be predictable and similar. This means... when I see `updateAutomaticTickets` in `LottoMachine`, I would expect `updateManualTickets` in `LottoMachine` as well.

Manual lotto being tightly coupled to user input hence putting it in the controller might be differed from person to person, and even differ between the past self and present self. That's why I would follow the past structure, but remember, you asked for my opinion, and this is just my opinion 😉 

It'll be fun to talk about this with others as well.

pkeugine (changes requested):
Hello Sebas, nice work on step 2,
and honestly, you've refactored the project VERY cleanly that
it was hard for me to find places to comment on. 😅 

I still managed to pull it off, so it'll be nice to check them sometime.
I left the PR as request changes since you might want to change something.
If you feel like it can be merged (I mean, I think it's enough already)
hit me with a review request and we'll finish mission 1.

Good luck with Blackjack mission, and as always,
Happy hacking ♠️ ♠️ ♠️

README.md
SebasNadu: Thanks! Yes I totally agree, I intentionally kept the commits small to capture each step of the process. But it definitely added some noise and made it harder to follow.

I'll definitely take a deeper look at rebase, squash, and fixup.. Appreciate the tip! 💪

src/main/kotlin/lotto/domain/PurchaseSession.kt
SebasNadu: Yes! Totally, that make it completely more readable. Thanks!

src/test/kotlin/lotto/domain/PurchaseSessionTest.kt
SebasNadu: That's amazing, thanks for the tip! With Kotlin and JUnit, there are so many options that it can sometimes feel a bit overwhelming😅

src/main/kotlin/lotto/domain/ResultAnalyzer.kt
SebasNadu: Yes, I felt the same way. I wanted to experiment with the concept of immutability, but I also had the sense that it added unnecessary complexity. My initial goal was to create a central place to manage session data in case the app was extended to support multiple sessions.

While immutability does provide a “safe” way to handle updates, I’m starting to think the cost of constantly creating new instances isn’t justified at this level of complexity, it has more sense in your example where you need to share a state safety, but this is not the real case in my implementation.

I also agree that PurchaseSession ended up with too much responsibility. It made sense at first, but now it feels a bit bloated.

Thanks again for your insights 🙏 They’ve been really helpful in reevaluating the direction.

src/test/kotlin/lotto/domain/PurchaseSessionTest.kt
SebasNadu: You're right!! Thanks so much, I misunderstood how .copy() works!

Even though PurchaseSession is immutable, .copy() performs a shallow copy, not a true deep copy 😮. Thanks again for pointing out these small details, they really help me clarify key concepts!

src/main/kotlin/lotto/controller/LottoController.kt
SebasNadu: Thanks for sharing with me your take, and I agree with your 'simple-is-best' philosophy, next time I will try to follow this line and prioritize simplicity and predictable logic placements. I really appreciate your opinion and in general It was really nice to have these conversations with you. Thank you!

SebasNadu:
Thank you so much for your feedback, PK🙌! You pay so much attention to detail, always bring thoughtful questions and great tips, and really push me to reflect and deepen my understanding. I enjoy receiving your feedback, and I hope we can keep having these chats in the future🙏.

Best,
Sebastian

pkeugine (approved):
Hello Sebas,
A lot has happened in a week, huh?
I hope going through this mission with me was a fun experience.

We'll wrap up Lotto mission here,
and I also look forward to meeting you again.

Happy hacking! 👊👊👊

PR #25, opened 2025-07-07
[Lotto - Step2] Submit Vito's mission

VittorioDeMarzi:
Hi Reviewer,

This is my pull request for Lotto Step 2.

After some refactoring based on the feedback from Step 1, I worked on the new features required for this step.

NEW FEATURES
 - I added the option for the user to enter Lotto numbers manually, in addition to the automatically generated ones. 
 - The user can now:
    -  input the purchase amount 
    - decide how many tickets they want to create manually
    - enter the numbers for each of those tickets.
- I also introduced a LottoNumber class as a wrapper for a single Int, following the First-Class Collection principle, as suggested in the requirements.
- The manual ticket numbers are validated to make sure each one has exactly 6 unique numbers in the range from 1 to 45. - There's also validation on the number of manual tickets requested, to ensure it's a positive number and doesn't exceed the total number of tickets allowed based on the purchase amount.

At this stage, I still plan to improve the structure of the LottoController, possibly by introducing a LottoService class to better separate responsibilities and reduce dependencies. I’m curious to hear your thoughts on this, and if you have any suggestions that could help me improve the overall design.

Lastly, I’m aware that there is still some duplicated validation logic across different layers. I plan to clean that up after your review or during the next step, by removing unnecessary duplication and making the structure more consistent.

Thanks again for your time and feedback, I really appreciate it!


README.md
```diff
@@ -6,11 +6,23 @@
...
+    - [x] number must lower than number of tickets
+- [x] calculate number of ticket user want to manually generate
+- [x] calculate number of tickets the machine has to automatically generate
+- [x] Display number of manual and automatic tickets
+- [x] Input and Validation numbers for manual tickets:
+    - [x] each ticket has 6 numbers
+    - [x] Numbers must be in the range 1 to 45
+
+- [ ] Generate automatic LottoTickets
     - [x] each ticket has 6 numbers
```
Hyeon9mak, line 21: 😉

src/main/kotlin/lotto/controller/LottoController.kt
```diff
@@ -14,29 +14,31 @@ class LottoController(
...
+        val result = winningLogic.determineWinningTickets(tickets, winningLotto)
+        val profitRate = result.calculateProfitRate()
+        resultView.displayWinningStatistics(result, profitRate)
     }
 
     fun handlePurchase(purchaseAmount: Int): List<Lotto> {
         val numberOfTickets = ticketFactory.calculateNumberOfTickets(purchaseAmount)
-        val tickets = ticketFactory.generateTickets(numberOfTickets)
-        resultView.displayNumberOfTickets(numberOfTickets)
+        val manualTicketsNumber = inputView.getManualTicketsNumber(numberOfTickets)
```
Hyeon9mak: It would be great if you could utilize the hints that IntelliJ provides to organize your code better. 😉 

image

src/main/kotlin/lotto/model/Lotto.kt
```diff
@@ -2,23 +2,26 @@ package lotto.model
...
-    fun containsBonus(bonusNumber: Int): Boolean {
+    fun containsBonus(bonusNumber: LottoNumber): Boolean {
         return numbers.contains(bonusNumber)
     }
 
-    fun displayLotto(): String {
-        return numbers.joinToString(",", prefix = "[", postfix = "]")
+    fun getLottoNumbers(): Set<LottoNumber> {
+        return numbers
+    }
```
Hyeon9mak: how about this?
it generate same getter method in Java after build the Kotlin code.

```kotlin
data class Lotto(val numbers: Set<LottoNumber>) {
    // ...
    // fun getLottoNumbers(): Set<LottoNumber> {
    //    return numbers
    // }
}
```

src/main/kotlin/lotto/model/LottoNumber.kt
```diff
@@ -0,0 +1,18 @@
+package lotto.model
+
+data class LottoNumber(private val value: Int) {
```
Hyeon9mak: what's different between value class?
https://kotlinlang.org/docs/inline-classes.html
https://kotlinlang.org/docs/data-classes.html

src/main/kotlin/lotto/model/Random.kt
```diff
@@ -1,7 +1,7 @@
 package lotto.model
 
 object Random {
-    fun generateListOfSixRandomNumbers(): List<Int> {
-        return (1..50).shuffled().take(6)
+    fun generateSetOfSixRandomNumbers(): Set<Int> {
+        return (1..50).shuffled().take(6).sorted().toSet()
```
Hyeon9mak, line 5: detail 👍

src/main/kotlin/lotto/model/TicketFactory.kt
```diff
@@ -9,12 +9,22 @@ object TicketFactory {
...
+
+    fun generateAutomaticTickets(numberOfTickets: Int): List<Lotto> {
+        return List(numberOfTickets) { Lotto.fromInts(Random.generateSetOfSixRandomNumbers()) }
+    }
+
+    fun generateManualTickets(manualTicketsList: List<Set<Int>>): List<Lotto> {
+        return manualTicketsList
+            .map { ticket -> Lotto.fromInts(ticket) }
+            .toList()
     }
```
Hyeon9mak: simple & nice code.
but those function(method)s are public for test code.

let's think about.
If it's really important logic that absolutely needs to be tested,
but it's difficult to test because it's private,
isn't that perhaps a signal that it should be separated into a separate object?

src/test/kotlin/lotto/core/InputValidationTest.kt
```diff
@@ -1,38 +1,120 @@
...
+        val purchaseAmount = 999
+        assertThrows<IllegalArgumentException> { InputValidation.validatePurchaseAmount(purchaseAmount) }
     }
 
     @Test
-    fun `winning numbers should be unique numbers`() {
+    fun `should throw if negative number`() {
+        val purchaseAmount = 0
+        assertThrows<IllegalArgumentException> { InputValidation.validatePurchaseAmount(purchaseAmount) }
+    }
```
Hyeon9mak: not important, but 0 is not negative 😉.
how about this message?
`purcahse must be positive number`

src/test/kotlin/lotto/core/InputValidationTest.kt
```diff
@@ -1,38 +1,120 @@
...
+    @Test
+    fun `should throw if bonus number in the winning numbers`() {
+        val bonusNumber = 1
+        val winningNumber = setOf(1, 2, 3, 4, 5, 6)
+        assertThrows<IllegalArgumentException> {
+            InputValidation.validateBonusNumberInput(
+                winningNumbers = winningNumber,
+                bonusNumberInput = bonusNumber,
+            )
+        }
```
Hyeon9mak: nice & detail test cases.
what happend in bonusNumber 0?

Hyeon9mak (changes requested):
Great work in step2 Vito!

> I still plan to improve the structure of the LottoController, possibly by introducing a LottoService class to better separate responsibilities and reduce dependencies. I’m curious to hear your thoughts on this, and if you have any suggestions that could help me improve the overall design.

It's excellent that you're thinking about various ways to reduce dependencies in the `Controller`. 👏 👏 👏 

However, I have doubts about whether it necessarily has to be a `Service`.
I think `Service` is something with a broad name that can do anything, like `Util` or `Factory` - toooooooooooooo common.

What do you think the `Service` is, Vito?
If you want to represent another object (not `Controller`) that helps cooperation between domain models, how about consciously using a different name, like `LottoGame`?

I want to encourage Vito to develop a habit of not unconditionally using the name `Service`. 😉 

Left some easy-comments on this PR.
Have any questions, plz feel free to leave a comment. 😄

src/main/kotlin/lotto/model/Lotto.kt
VittorioDeMarzi: Instead of exposing the property how about this?

`val lottoNumbers`
 `       get() = numbers`

src/main/kotlin/lotto/model/LottoNumber.kt
VittorioDeMarzi: 1. Data Class (data class LottoNumber)
Creates a full-fledged object, with:
 - equals(), hashCode(), toString()
 - copy() function
 - Extra memory allocation
 - Has runtime overhead: each instance is a separate object in memory.

2. Value Class (value class LottoNumber)
 - Introduced in Kotlin to avoid memory overhead for wrapper types.
 - The JVM treats it like the underlying value (in this case, an Int) at runtime, when possible.
 - No copy() or componentN() methods.
 - More performant, especially when used in collections or tight loops.
 - Only one property allowed.
 
 In this case 

`@JvmInline`
`value class LottoNumber(private val value: Int)`

It’s more efficient and semantically clearer. I will refactor it

src/main/kotlin/lotto/model/TicketFactory.kt
VittorioDeMarzi: You're right, these helper methods (generateAutomaticTickets, generateManualTickets) are currently public mostly to make them testable.

I've been thinking about that too:
if the logic inside them is important enough to deserve its own tests, maybe it's a sign that it belongs in a separate class, like a TicketGenerator, where generation responsibilities are better encapsulated.

VittorioDeMarzi:
Thank you so much for the thoughtful review and all the helpful suggestions! 🙏

I've implemented the changes you recommended, including improving naming, simplifying logic, and addressing visibility concerns around helper methods. Your feedback about the overuse of the term Service really made me reflect. You're absolutely right: using a more expressive name like LottoGame can better communicate intent and avoid generic abstractions. I’ll keep that in mind moving forward. 

Before diving deeper into a full refactor or introducing a LottoGame class, I’d like to shift focus temporarily to refactoring the Blackjack project, which I expect will take quite a bit of time and mental bandwidth.

I hope the changes I’ve made here are sufficient for now, but if you feel there are still responsibilities that should be further separated or architectural improvements to be made, please let me know, I’m happy to revisit it! 😄

Thanks again for the great guidance!

src/main/kotlin/lotto/model/Lotto.kt
Hyeon9mak: Do we really need to add an extra getter for this? 🤔 
suggested code would result in two separate getter methods when converted to Java.

src/main/kotlin/lotto/model/LottoNumber.kt
```diff
@@ -0,0 +1,19 @@
+package lotto.model
+
+@JvmInline
+value class LottoNumber(private val value: Int) {
+    companion object {
+        private const val MINIMUM_NUMBER = 1
```
Hyeon9mak, line 6: 😉 👍

src/main/kotlin/lotto/model/TicketGenerator.kt
```diff
@@ -0,0 +1,15 @@
...
+        return List(count) {
+            Lotto.fromInts(Random.generateSetOfSixRandomNumbers())
+        }
+    }
+
+    fun generateManualTickets(inputs: List<Set<Int>>): List<Lotto> {
+        return inputs
+            .map { ticket -> Lotto.fromInts(ticket) }
+            .toList()
+    }
```
Hyeon9mak, line 14: Nice!

src/test/kotlin/lotto/core/InputValidationTest.kt
```diff
@@ -1,38 +1,138 @@
...
+    @Test
+    fun `should throw negative`() {
+        val purchaseAmount = -1000
+        assertThrows<IllegalArgumentException> { InputValidation.validatePurchaseAmount(purchaseAmount) }
+    }
+
+    @Test
+    fun `should throw if not divisible by 1000`() {
+        val purchaseAmount = 1500
+        assertThrows<IllegalArgumentException> { InputValidation.validatePurchaseAmount(purchaseAmount) }
```
Hyeon9mak, line 31: 😉 👍

Hyeon9mak (approved):
Hi Vitto!
Great job seeing the Lotto mission through to the end!

All the requirements have been met, and you've addressed the feedback well.
So I'm approving this PR. 😄 

I hope you can focus fully on the Blackjack mission now
and continue to grow even more!

PR #26, opened 2025-07-07
[Lotto - Step 2] Submit codebrew42's mission

jieun-pak:
I misunderstood the procedure of submission and due to time constraints I couldn’t apply TDD properly this time. However, I tried my best to implement all the required features as best as I could.

I’d appreciate feedback on:
1.  Kotlin syntax violations
2. any other critical logical errors.

Thank you so much in advance!

-Jin Park


src/main/kotlin/controller/Controller.kt
```diff
@@ -2,6 +2,7 @@ package controller
```
pkeugine, line 2: I see that you've been working on step 2 as we were reviewing step 1. 👏 
That resulted in kind of a complex commit log.
Once you get some time, I recommend trying out `rebase` and `cherry-pick` features.
They are implemented in IntelliJ, so definitely check it out; it'll be easier to use than command line, especially when cherry-picking.

Oh and these features are used daily in actual work, so they're pretty important as well. :)

src/main/kotlin/controller/Controller.kt
```diff
@@ -11,38 +12,71 @@ class Controller {
         inputView: InputView,
         outputView: OutputView,
     ) {
-        val lotto = handleLottoPurchase(inputView, outputView)
+        val lotto = setNumbersOfEachTickets(inputView, outputView)
+        fillNumbersToEachTickets(lotto, inputView, outputView)
+        outputView.displayNumberOfLottoTickets(lotto)
+        outputView.displayTickets(lotto)
```
pkeugine: I see there's an `outputview` used twice in a row.
Why don't we make it so that one method handles it all?
When we need to separate these displays into two for logical or graphical reason, that's when dividing them could be great.

src/main/kotlin/view/OutputView.kt
```diff
@@ -7,8 +7,21 @@ class OutputView {
...
+            "Purchased ${lotto.numberOfManualTickets} manual and ${lotto.numberOfTotalTickets} automatic tickets.",
+        )
+    }
+
+    fun displayTicketsWithoutBrackets(lotto: model.Lotto) {
+        for (ticket in lotto.tickets) {
+            println(ticket.numbers.sorted().joinToString(", "))
+        }
+        println()
     }
```
pkeugine: This method is not used in any location. Let's remove it :)

src/main/kotlin/view/OutputView.kt
```diff
@@ -18,6 +31,10 @@ class OutputView {
         println()
     }
 
+    fun displayTicketWithoutBrackets(ticket: model.Ticket) {
+        println(ticket.numbers.sorted().joinToString(", "))
+    }
+
```
pkeugine, line 42: I see some places where we can apply [this comment.](https://github.com/herotechcourse/kotlin-lotto/pull/2#discussion_r2190373850)

src/main/kotlin/controller/Controller.kt
```diff
@@ -11,38 +12,71 @@ class Controller {
...
+                lotto.numberOfAutomaticTickets = lotto.numberOfTotalTickets - manualCount
+                return lotto
             } catch (e: IllegalArgumentException) {
-                println("${e.message}")
+                println(e.message ?: "Unknown error")
                 lastException = e
             }
         }
-        throw lastException!!
+        throw lastException ?: IllegalArgumentException("Unknown error")
```
pkeugine: Instead of keeping the exception for later, why don't we just :
```Kotlin
    throw IllegalArgumentException("Too many failed attempts. Please restart the program.")
```

Also, I think we can utilize `repeat(maxAttempts)` here.

src/main/kotlin/controller/Controller.kt
```diff
@@ -11,38 +12,71 @@ class Controller {
...
+        inputView: InputView,
+        outputView: OutputView,
+    ) {
+        generateManualTickets(lotto, inputView)
+        if (lotto.numberOfManualTickets > 0) {
+            for (i in 0 until lotto.numberOfManualTickets) {
+                outputView.displayTicketWithoutBrackets(lotto.tickets[i])
+            }
+            println()
+        }
```
pkeugine: I think we can put this part in `OutputView` especially since `println()` sneaked out into the controller. 😂 

Let's look at other places in this controller and see if we can move the input, output part into views.

src/main/kotlin/model/Lotto.kt
```diff
@@ -2,22 +2,29 @@ package model
...
+    fun generateAutomaticTickets() {
+        numberOfAutomaticTickets = numberOfTotalTickets - numberOfManualTickets
+        for (i in 0 until numberOfAutomaticTickets) {
             tickets.add(createTicket())
         }
     }
 
+    fun createTicketForController(): Ticket {
+        return createTicket()
+    }
```
pkeugine: This method is wrapping a private method, without any other logic in it.
Why don't we just make `createTicket` public?
If there's a reason behind this implementation, feel free to tell me about it :)

src/main/kotlin/controller/Controller.kt
```diff
@@ -11,38 +12,71 @@ class Controller {
...
+        val maxAttempts = 3
         var lastException: IllegalArgumentException? = null
         for (i in 1..maxAttempts) {
             try {
                 val purchaseAmount = inputView.getPurchaseAmount()
-                return Lotto(purchaseAmount)
+                val lotto = Lotto(purchaseAmount)
+                outputView.displayPurchaseAmount(lotto)
+                val manualCount = inputView.getNumberOfManualTickets()
+                lotto.numberOfManualTickets = manualCount
```
pkeugine: By leaving `numberOfManualTickets` as `var` and using this logic, we're actually using a `setter` to fill in the value in `Lotto`. Although this seems simple and easy, it's not really something I would recommend when creating an object. It'll be nice if you could look up the reason behind it!

**Hint** : When we talk about setters and changing values, it always comes down to **immutability**.

src/main/kotlin/view/InputView.kt
```diff
@@ -41,10 +43,35 @@ class InputView {
...
+            inputAsString.split(delimiter)
+                .map { it.trim() }
+                .mapNotNull { it.toIntOrNull() }
+        return if (inputList.isNotEmpty() && inputList.all { it.toString() in inputAsString.split(delimiter).map { s -> s.trim() } }) {
+            inputList
+        } else {
+            println(ErrorMessages.INPUT_INVALID_DIGITS.message)
+            return getUserInputAsListOfInt(prompt, numberOfAttempts - 1, delimiter)
+        }
+    }
```
pkeugine: As we progressed into the mission, the code feels like it's been rushed.
I think this method could be refactored to be more readable,
and with unnecessary logic removed.
Why don't we give it a try?

src/main/kotlin/view/InputView.kt
```diff
@@ -41,10 +43,35 @@ class InputView {
         }
     }
 
+    fun getUserInputAsListOfInt(
```
pkeugine, line 37: Please check the [previous PR reviews such as this one](https://github.com/herotechcourse/kotlin-lotto/pull/2#discussion_r2190370793), and see if they were applied.
If not, it'll be nice to see the reason behind it. I'm open to any kind of opinions.

pkeugine (changes requested):
Hello Jin,
This PR looks simple and it was very easy to understand.
I've added some comments, and I strongly recommend you check it out.
I'll leave this PR as a 'request changes' so once you're done,
feel free to ask for another review.

If you have any questions or thoughts, comment in this PR or send me a DM anytime.
Happy hacking 🔥 🔥 🔥

pkeugine (changes requested):
Hello Jin, since this PR is marked as `requested for review`, I'll keep it as `request changes` for now. Feel free to work on this mission when you have the time. Cheers!

src/main/kotlin/controller/Controller.kt
jieun-pak: Actually I learned about how to use them a few days ago, I think I need more practice and thank you for reminding me about it!

jieun-pak:
Q. My previous code seems fine, so I was wondering if there's a specific reason to prefer the multiline string in this case other than scalability? I feel the original version might be slightly more concise and readable in this small project when displaying a single line of output. I appreciate your opinion on it :)


previous code
```
    fun displayTicketWithoutBrackets(ticket: model.Ticket) {
        println("Ticket Numbers: ${ticket.numbers.sorted().joinToString(", ")}")
    }
```


improved one: multiple line string is recommended
```
 fun displayTicketWithoutBrackets(ticket: model.Ticket) {
        val formattedNumbers = ticket.numbers.sorted().joinToString(", ")
        println("""
            Ticket Numbers: $formattedNumbers
        """.trimIndent())
    }
```

src/main/kotlin/controller/Controller.kt
jieun-pak: I can't agree more! thanks.

src/main/kotlin/controller/Controller.kt
jieun-pak: I implemented it, I'll keep in mind!

src/main/kotlin/view/InputView.kt
jieun-pak: I didn't mean to overlook your suggestion. It's just a habit I have when coding alone, I'll keep it in mind going forward!

src/main/kotlin/controller/Controller.kt
jieun-pak: I implemented it, thank you!

src/main/kotlin/model/Lotto.kt
jieun-pak: I can't agree more, there's no reason behind it... I'll care more about these type of unnecessary methods, thanks!

src/main/kotlin/view/InputView.kt
jieun-pak: Now I see it's so messy. I tried to improve it, and now this part seems much cleaner, thank you for pointing it out!

src/main/kotlin/controller/Controller.kt
jieun-pak: You're absolutely right that using a setter approach (by keeping numberOfManualTickets as var) violates immutability principles. I learned that when objects can be modified after creation, it leads to unpredictable state changes.

Before
```
class Lotto(val purchaseAmount: Int) {
    val numberOfTotalTickets: Int
    var numberOfManualTickets: Int = 0
...
}

lotto.numberOfManualTickets = manualCount   // Mutating after creation

```
  After (Improved):
```
  val lotto = Lotto(purchaseAmount, manualCount)  // Complete object from start
```

I appreciate your detailed feedback, thank you!

src/main/kotlin/view/OutputView.kt
```diff
@@ -33,16 +55,28 @@ class OutputView {
...
+            """
+            Winning Statistics
+            ------------------
+            3 Matches (5,000 KRW) - ${results[4]} tickets
+            4 Matches (50,000 KRW) - ${results[3]}  tickets
+            5 Matches (1,500,000 KRW) - ${results[2]}  tickets
+            5 Matches + Bonus Ball (30,000,000 KRW) - ${results[1]}  tickets
+            6 Matches (2,000,000,000 KRW) - ${results[0]}  tickets
+            """.trimIndent(),
+        )
```
pkeugine, line 68: **Q. My previous code seems fine, so I was wondering if there's a specific reason to prefer the multiline string in this case other than scalability?**

It's my mistake on this part.
Right after leaving the comment about `displayTicketsWithoutBrackets` not being used, I guess my hand thought that method was also `displayMatchResults`. 😂

I agree with you completely, and I don't feel the need of multi-line string when printing out one line output. Let this comment pass by, and sorry for making you think about this.

src/main/kotlin/view/InputView.kt
```diff
@@ -4,91 +4,169 @@ import model.Lotto
...
+                    .filter { it.isNotEmpty() }
+                    .mapNotNull { it.toIntOrNull() }
+            if (result.isNotEmpty()) {
+                return result
+            }
             println(ErrorMessages.INPUT_INVALID_DIGITS.message)
-            return getUserInputAsInt(prompt, numberOfAttempts - 1)
         }
+        throw IllegalArgumentException(ErrorMessages.INPUT_TOO_MANY_ATTEMPT.message)
     }
```
pkeugine, line 55: This method could be private.
There's also another one in `OutputView` where we can change to private, so I think it'll be good to check these parts when refactoring :)

src/test/kotlin/model/LottoTest.kt
```diff
@@ -9,88 +9,88 @@ class LottoTest {
...
+            Ticket.fromInts(listOf(1, 2, 3, 4, 5))
         }
     }
 
     @Test
     fun `should throw exception - a ticket with duplicates`() {
         assertThrows<IllegalArgumentException> {
-            Ticket(listOf(1, 1, 2, 3, 4, 5))
+            Ticket.fromInts(listOf(1, 1, 2, 3, 4, 5))
         }
```
pkeugine, line 80: It would be great to let these `Ticket` related tests to be maintained in `TicketTest`, since that would be a place where we would be looking for.

pkeugine (approved):
Hello Jin, and I'm glad that we finally get to finish this mission! 😄 
Nice work on keeping this mission going, and I see you've refactored this project very well.

I'll merge this PR, and I hope you do well in current mission!
Happy Hacking! 🔥 🔥 🔥

src/main/kotlin/view/OutputView.kt
jieun-pak: Never mind, it helped me to think about this topic, it was helpful! thank you anyway!

src/main/kotlin/view/InputView.kt
jieun-pak: I misunderstood that if i want to test I shouldn't make private functions, and then realized that way of thinking might ruin the architecture which is not good. I'll always keep in mind to make functions private if possible and needed.
Thank you!

src/test/kotlin/model/LottoTest.kt
jieun-pak: I can't agree more, thanks !

jieun-pak:
I've finally applied most of your feedback in the new branch. 
I learned so much through your comments and questions. I appreciate it.
The more I look back at my code, the more embarrassed I feel 😅...
but I know I still have a long way to go and a lot to learn.
I'm choosing to see it positively and keep growing.

Thanks again!

PR #27, opened 2025-07-07
[Lotto - Step2] Submit Gabi's mission

knopgm:
Hi Eugine,
I'm sorry for the confuse code for these step2. 
I blocked and almost lost todays deadline.
I'm planning in refactor many things, but I'll be happy to follow your orientations again.

In this pull-request I created a feature that asks for the user if they want to input their own numbers.
InputView asks for quantity of tickets and the numbers, while LottoMachine create the manual tickets and treat it as a LottoTicket should be treated.

I'm open to improve my pull requests, if this message wasn't helpful too. 😅 
Thank you again for your time and the great last review you gave me! 🙏


src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,20 +1,19 @@
...
-    fun createTickets(userAmount: Int): List<Lotto> {
+    fun createGeneratedTickets(
+        totalPurchasedTickets: Int,
+        manualTicketsQuantity: Int,
+    ): List<Lotto> {
         val lottos = mutableListOf<Lotto>()
-        val amountOfTickets = calculateTickets(userAmount)
-        var count = 0
-        while (count < amountOfTickets) {
+        var ticketsToBeGenerated = totalPurchasedTickets - manualTicketsQuantity
```
pkeugine:
```suggestion
        val ticketsToBeGenerated = totalPurchasedTickets - manualTicketsQuantity
```

It'll be great to keep things `val` when possible. :)

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,20 +1,19 @@
...
+        var countGeneratedTickets = 0
+        while (countGeneratedTickets < ticketsToBeGenerated) {
             lottos.add(Lotto.create())
-            count++
+            countGeneratedTickets++
```
pkeugine, line 15:
```Kotlin
repeat(ticketsToBeGenerated) {
    lottos.add(Lotto.create())
}
```

Same applies to for, loop, or repeat operations as well.
We don't have to count unless there's a chance for we to modify it or get a different outcome.

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,20 +1,19 @@
...
         }
-        return lottos
-    }
 
-    private fun calculateTickets(userAmount: Int): Int {
-        val numberOfTickets = userAmount / LOTTO_PRICE
-        return numberOfTickets
+        return lottos
```
pkeugine:
```Kotlin
    val ticketsToBeGenerated = totalPurchasedTickets - manualTicketsQuantity
    return List(ticketsToBeGenerated) { Lotto.create() }
```

If we use Kotlin's power to it's fullest,
this logic can be simplified into above two lines.

Although making logic short isn't always the best option,
it's good to know that Kotlin offers these features for us to use!

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -55,6 +54,10 @@ class LottoMachine() {
         return totalPrize
     }
 
+    fun createManualTickets(userManualTicketNumbers: List<List<Int>>): List<Lotto> {
+        return userManualTicketNumbers.map { ticket -> Lotto(ticket.map { number -> LottoNumber(number) }) }
+    }
```
pkeugine, line 62: Like above comment, short one-liner might not always be the right option.
I tried tweaking this method to make it more readable such as :  
```Kotlin
    fun createManualTickets(userManualTicketNumbers: List<List<Int>>): List<Lotto> {
        return userManualTicketNumbers.map { numberList ->
            val lottoNumbers = numberList.map { number -> LottoNumber(number) }
            Lotto(lottoNumbers)
        }
    }
```
but some might disagree 🤔 .

If we move a part of this to a static factory method in `Lotto` such as :  
```Kotlin
        fun fromInteger(numbers: List<Int>): Lotto {
            val lottoNumbers = numbers.map { number -> LottoNumber(number) }
            return Lotto(lottoNumbers)
        }
```

then we'll be able to simplify it more by making it like this :  
```Kotlin
    fun createManualTickets(userManualTicketNumbers: List<List<Int>>): List<Lotto> {
        return userManualTicketNumbers.map(Lotto::fromInteger)
    }
```

Try and see what looks nice to you. It's pretty fun to tweak with logic and find out what you see the most when you think about 'readability'.

src/main/kotlin/lotto/LottoNumber.kt
```diff
@@ -1,11 +1,15 @@
...
     init {
         require(isInRange(value))
     }
 
     private fun isInRange(value: Int): Boolean {
         return value in 1..45
     }
+
+    override fun compareTo(other: LottoNumber): Int {
+        return value - other.value
```
pkeugine, line 13:
```suggestion
        return value.compareTo(other.value)
```

What could be the difference? I mean, they both look like subtraction right?
Although this won't affect lotto number comparison, it does have a difference.
It's an interesting topic. 😋

src/main/kotlin/lotto/Application.kt
```diff
@@ -3,10 +3,18 @@ package lotto
 fun main() {
```
pkeugine, line 3: Why don't we check the comments from previous PR and see if we can apply some of them?
I see that the schedule seems pretty tight, so give it a try if it's possible, but don't stress it.

pkeugine (changes requested):
Hello Gabi, nice work on step 2.
I see that deadline came by very quickly, but you still managed to send me this PR 👏 
I've left some comments with suggestions.
It's not too hard, so I recommend checking it out.

If you have any questions or thoughts,
feel free to leave a comment or DM me anytime.
Happy hacking! 💪 💪 💪

src/main/kotlin/lotto/LottoMachine.kt
knopgm: Jason has presented me to repeat yesterday. I will use it from now own when I see it fits 🙏 😄

src/main/kotlin/lotto/InputValidator.kt
```diff
@@ -8,6 +8,27 @@ class InputValidator {
...
+        }
+    }
+
+    fun validateManualTicketsAmount(
+        ticketQuantity: String,
+        totalPurchasedTickets: Int,
+    ) {
+        if (ticketQuantity.toIntOrNull() == null) {
+            throw IllegalArgumentException("Input is not a number")
+        }
```
pkeugine, line 22:
```suggestion
        requireNotNull(ticketQuantity.toIntOrNull()) { "Input is not a number" }
```

These kind of validations can be done with `requireNotNull` :)

src/main/kotlin/lotto/InputValidator.kt
```diff
@@ -8,6 +8,27 @@ class InputValidator {
...
+    }
+
+    fun validateManualListOfNumbers(numbersList: String) {
+        if (numbersList.toIntOrNull() == null) {
+            throw IllegalArgumentException("Input is not a number")
+        }
     }
 
     fun validateWinningNumbers(userInput: List<String>) {
```
pkeugine, line 34: while trying out the application, I found a bug where winning numbers don't get parsed correctly when I give `1, 2, 3, 4, 5, 6` as an input. This is because `trim()` didn't get applied to individual numbers, and the program wasn't able to `toIntOrNull` it correctly. If you have the time, it'll be great to check this out and see why it didn't work!

pkeugine (approved):
Hello Gabi, and I'm glad we finally covered step 2!
Nice work 👍 

I've left some comments, and one of them actually includes an important part of this project so I highly recommend you check it out.

It was nice going through this mission with you,
and I hope you do well in other missions as well :)

As always,
Happy Hacking! 💪 💪 💪

PR #29, opened 2025-07-07
[Lotto - Step2] Submit Bojana's mission

bojana-petroska:
Hi GuMak!

I fully implemented the feedback from the first review. The second feedback is not yet fully implemented.

I implemented the feature that allows the user to manually add lotto numbers. The logic for it is in the `LottoMashine`. The `InputView` is handling all the user inputs, and the `ResultView` prints the output.

The main became very long, so it was my intention to refactor it (gladly would hear some tips here ; ) My concern is (not)following the separation of concern concept.

I would highly appreciate your Input!

Thank you very much!


README.md
```diff
@@ -18,12 +18,15 @@ Update the winning statistics to include the second prize category.
...
 - [x] generate 6 unique, random, numbers in range 1 to 45
-- [x] machine issues lotto tickets
+- [x] machine issues the rest of the lotto tickets(automated)
 - [x] figure out winning tickets
 
 #### Input
 - [x] get input of amount
+- [x] get the numbers for the manually created tickets
 - [x] get the winning numbers
 - [x] get the bonus number
```
Hyeon9mak, line 31: 😉 👍

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,13 +1,41 @@
...
 
-class LottoMachine(val money: Money) {
+class LottoMachine(val money: Money, manuallyEnteredNumbers: List<LottoNumbers>) {
     val tickets: List<Lotto>
 
     init {
-        val ticketCount: Int = money.amount / TICKET_COST
-        tickets = List(ticketCount) { Lotto(LottoNumbers.of(generateRandomNumber())) }
+        val maxTickets = money.amount / TICKET_COST
+        require(manuallyEnteredNumbers.size <= maxTickets) {
```
Hyeon9mak: How about ask divide result to `Money` class?
I think the `TICKET_COST` value used in most cases with `Money` class.

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,13 +1,41 @@
...
+            }
+        return manualLottoTickets
+    }
+
+    private fun getAutomaticRemainingTicketsCount(manualLottoTickets: List<Lotto>): Int {
+        val purchasedManualAmount = manualLottoTickets.size * TICKET_COST
+        val remainingMoney = money.amount - purchasedManualAmount
+        val remainingTicketCount: Int = remainingMoney / TICKET_COST
+        return remainingTicketCount
+    }
```
Hyeon9mak: If we move `TICKET_COUNT` to the `Money` class, this method will likely be affected.
But instead of checking the price again here, it might be better to simply use the ticket count to perform `getAutomaticRemainingTicketsCount`.

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,13 +1,41 @@
...
+        tickets = manualLottoTickets + automaticLottoTickets
+    }
+
+    private fun generateManualLottoTickets(numbers: List<LottoNumbers>): List<Lotto> {
+        val manualLottoTickets =
+            numbers.map { lottoNumbers ->
+                Lotto(lottoNumbers)
+            }
+        return manualLottoTickets
+    }
```
Hyeon9mak: How about this?

```suggestion
    private fun generateManualLottoTickets(numbers: List<LottoNumbers>): List<Lotto> {
        return numbers.map(::Lotto)
    }
```

src/main/kotlin/view/InputView.kt
```diff
@@ -1,22 +1,56 @@
...
+
+    fun getManualTicketCount(maxCount: Int): Int {
+        println("Enter the number of manual tickets to purchase.")
+        val inputString = readLine() ?: ""
+        val count = inputString.toIntOrNull()
+        require(count != null && count >= 0 && count <= maxCount) {
+            "Manual tickets must be between 0 and $maxCount."
+        }
+        return count
+    }
```
Hyeon9mak: The rule about not exceeding the maximum count,
seems more like something to discuss with the planner, not the designer.

In that case, I think it belongs in the domain layer, not the view! 🤔

src/main/kotlin/lotto/Application.kt
```diff
@@ -6,11 +6,23 @@ import view.ResultView
...
+        allManualTickets.add(singleManualTicketNumbers)
+    }
+
+    val lottoMachine = LottoMachine(purchaseAmount, allManualTickets)
     val lottoTickets = lottoMachine.tickets
 
+    val automatedTicketCount = lottoTickets.size - manuelTicketCount
     ResultView.viewTicketsAmount(lottoTickets.size)
-    ResultView.viewTickets(lottoTickets)
+    ResultView.viewTickets(lottoTickets, manuelTicketCount, automatedTicketCount)
```
Hyeon9mak: It seems that the `main` became lengthy mostly because,
you’re extracting values from objects using getters and assigning them to variables.

Instead of pulling out properties via getters and handling them externally, 
how about sending messages to the objects to get the information you need?

for example,

```kotlin
val lottoMachine = LottoMachine(purchaseAmount, allManualTickets)
ResultView.viewTicketsAmount(maxTickets)
ResultView.viewTickets(lottoMachine.tickets, manuelTicketCount, lottoMachine.calculateAutoTicketCount)
```

src/main/kotlin/view/ResultView.kt
```diff
@@ -9,7 +9,12 @@ object ResultView {
         println("You have purchased $ticketsAmount tickets")
     }
 
-    fun viewTickets(tickets: List<Lotto>) {
+    fun viewTickets(
+        tickets: List<Lotto>,
+        manuelTicketsNumber: Int,
+        automaticTicketsNumber: Int,
+    ) {
+        println("Purchased $manuelTicketsNumber manual and $automaticTicketsNumber automatic tickets.")
```
Hyeon9mak, line 17: image

Not important, but how about formatting it to show only two decimal under floating point?
it makes more good UX!

Hyeon9mak (changes requested):
Hi Bojana, great work in step2!

I left some comments for improve `main` function visibility.
And including suggestions about moving constant values.

Have any questions, plz feel free to leave a comment. 😄

src/main/kotlin/lotto/LottoMachine.kt
bojana-petroska: great tip! I removed `TICKET_COST` and refactored a bit.

src/main/kotlin/lotto/Application.kt
bojana-petroska: I really wanted to do some refactoring here because it was not very readable. So I did some refactoring where I separated the `Purchase` and the `GameController` into helper objects, whereas I made a data class for the `PurchaseAmount` so I can return structure data like `money`, `tickets` and `manual count`.

src/main/kotlin/lotto/LottoController.kt
```diff
@@ -0,0 +1,39 @@
...
+object LottoController {
+    fun start() {
+        val purchaseResults = PurchaseManager.purchaseTickets()
+        showPurchasedTickets(purchaseResults)
+
+        val (winningNumbers, bonusNumber) = getWinningNumbersAndBonus()
+
+        val game = LottoGame(purchaseResults.tickets, Lotto(winningNumbers), bonusNumber)
+
+        showResults(game, purchaseResults.money.amount)
```
Hyeon9mak: What’s the difference between the `Controller` and the `Manager`?
Both seem to interact with the view to retrieve something, but I’m not quite sure how their responsibilities differ.
(My guess is that Manager might be a replacement for something like a Service?)

Could you share a bit more context on their intended roles?

src/main/kotlin/lotto/Money.kt
```diff
@@ -13,4 +15,13 @@ value class Money private constructor(val amount: Int) {
...
     }
+
+    fun canPurchaseTickets(ticketCount: Int) {
+        val max = maxTickets(amount)
+        require(ticketCount <= max) {
+            "You can't purchase more than $max tickets."
+        }
+    }
+
+    fun maxTickets(amount: Int): Int = amount / TICKET_COST
```
Hyeon9mak, line 26: 😉 👍 
Now u can generate test code much easier.

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,13 +1,29 @@
 package lotto
 
-private const val TICKET_COST = 1000
-
-class LottoMachine(val money: Money) {
+class LottoMachine(val money: Money, manuallyEnteredNumbers: List<LottoNumbers>) {
```
Hyeon9mak:
```suggestion
class LottoMachine(private val  money: Money, manuallyEnteredNumbers: List<LottoNumbers>) {
```

src/main/kotlin/lotto/LottoController.kt
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    private fun showPurchasedTickets(purchaseResults: PurchaseResult) {
+        val manualCount = purchaseResults.manualCount
+        val autoCount = purchaseResults.tickets.size - manualCount
```
Hyeon9mak: > Don't use getter. Send message to Object.

Maybe you could try some different ways to practice sticking to the rules?

```kotlin
val (manualCount, autoCount) = purchaseResults.ticketCounts()
```

Hyeon9mak (changes requested):
Hello Bojana!

I’ve left a few small comments, 
along with a question about the difference between the `Controller` and the `Manager`.

Since there aren’t many changes needed in the code itself,
I think it’d be helpful to focus on explaining the intended distinction between those two roles. 😉 

If you have any other questions, feel free to ask anytime.
Hope you have a nice weekend!

src/main/kotlin/lotto/LottoController.kt
bojana-petroska: Thank you for pointing this out!

Yes, indeed, my initial structure was structurally unclear. After reviewing your feedback and reading more about MVC and OOP, I realized:
- the `PurchaseManager` was not named properly and it acted like a controller and service.
- both the `LottoController` & `PurchaseManager` were interacting with the view and the game logic, which violated clear separation of concerns.

So, I did the following changes:
- I renamed `PurchaseManager` to `PurchaseService`, and focuses only on the business logic: purchasing tickets, validating amounts, building results.
- I moved all input/output responsibilities into the Controller, so it only handles the coordination: collecting input, invoking services, and rendering output.
- I removed the `Controller` & `Service` from the model package and put them in the controller package.
- the Model holds the state (like `PurchaseResult`, `Money`, `Lotto`), and encapsulates domain behavior.

I tried to have the new structure to follow the MVC pattern.

So, after reviewing this part of the code I made a conclusion in the way MVC should be applied. 
- service for business logic and independent from views.
- controllers should be the only ones interacting with the UI and coordinating the flow.
- good naming helps clarify intention and structure.

Thank you so much for this question, it made some things more clear to me : )

src/main/kotlin/lotto/PurchaseResult.kt
```diff
@@ -0,0 +1,26 @@
+package lotto
+
+import view.ResultView
+
+data class PurchaseResult(
+    val money: Money,
+    val tickets: List<Lotto>,
+    val manualCount: Int,
+) {
+    fun totalTicketCount(): Int = tickets.size
```
Hyeon9mak, line 10: `PurchaseResult` also seems to be a class that contains business logic,
but it's currently directly depending on the View. 😱 
It might be better to invert the dependency so that the `View` depends on `PurchaseResult` instead.

src/main/kotlin/controller/PurchaseService.kt
```diff
@@ -0,0 +1,24 @@
...
+        money.canPurchaseTickets(manualTickets.size)
+
+        val lottoMachine = LottoMachine(money, manualTickets)
+
+        return PurchaseResult(
+            money = money,
+            tickets = lottoMachine.provideTickets(),
+            manualCount = manualTickets.size,
+        )
+    }
```
Hyeon9mak, line 23: It's always ideal when the LottoController can stay out of the way and the game still progresses smoothly — and I think you did a good job refactoring toward that.

Taking it a step further, names like manager or service often carry overly broad responsibilities.
They can easily become catch-alls, which makes it tiring to track them down every time the requirements change.

Instead of relying on generic names like service, factory, manager, or utils out of habit, how about choosing names that are more descriptive and intuitive?
If the name refers to something concrete in the real world, it becomes much easier to understand and reason about.

And that’s what we call OOP. 😄

src/main/kotlin/lotto/Money.kt
Hyeon9mak: Actually, this comment means "There is no test code. u need test this!" 😉

Hyeon9mak (approved):
Hi Bojana!

Nice job quickly understanding the intent and making the change!
It’s great to see that you have a solid grasp of what the Controller’s role should be.

As you continue with the upcoming missions, try to keep your business logic decoupled from the Controller — that way, it’ll be much easier to switch from a terminal console text game to a web-based image game later on, since you’ll only need to adjust the view and controller layers. 👍 

All the requirements have been met, and you've addressed the feedback well — so I'm approving this PR.
Hope you enjoy the Blackjack mission too! 😄

PR #30, opened 2025-07-08
[Lotto Step 2] submit saragrosser's mission 

saragrosser:
Hi!  This PR implements the features required for Step 2 of the Lotto mission. The goal was to allow users to manually input their own Lotto numbers, and to clean up the codebase by removing unnecessary internal state from key classes.

What I added
* User can now choose how many manual tickets to enter
* Program asks for and validates 6 numbers per ticket
* Remaining tickets are generated automatically
* Summary is printed with count of manual and auto tickets
* Winning numbers are compared and statistics displayed as before

Refactoring / Clean-up
* LottoPrinter is now a stateless object with a pure generation method
* LottoMachine is now a pure class with a single function
* Removed all internal mutable state 
* Old unused method in ResultView deleted

Could you please provide feedback in the approaches I took to implement the new manual logic thank you!

### DIFF
--- a/src/main/kotlin/lotto/BonusNumber.kt
+++ b/src/main/kotlin/lotto/BonusNumber.kt
@@ -1,0 +1,4 @@
+package lotto
+
+@JvmInline
+value class BonusNumber(val value: Int)
--- a/src/main/kotlin/lotto/BonusNumber.kt
+++ b/src/main/kotlin/lotto/BonusNumber.kt
@@ -1,4 +1,3 @@
 package lotto
 
-@JvmInline
-value class BonusNumber(val value: Int)
+data class BonusNumber(val value: Int)
--- a/src/main/kotlin/lotto/Lotto.kt
+++ b/src/main/kotlin/lotto/Lotto.kt
@@ -1,6 +1,6 @@
 package lotto
 
-object Lotto {
+class Lotto {
     fun run() {
         val amountOfMoney = retryable { InputView.getPurchaseAmount() }
         val totalTickets = amountOfMoney / LottoTicket.COST_OF_TICKET
@@ -10,8 +10,7 @@ object Lotto {
 
         val autoCount = totalTickets - manualCount
 
-        val lottoPrinter = LottoPrinter()
-        val autoTickets = lottoPrinter.generateLottoTickets(autoCount)
+        val autoTickets = LottoPrinter().generateLottoTickets(autoCount)
 
         val allTickets = manualTickets + autoTickets
 
@@ -21,11 +20,12 @@ object Lotto {
         val winningNumbers = retryable { InputView.getLastWeekWinningNumbers() }
         val bonusNumber = retryable { InputView.getBonusNumber() }
 
-        val machine: LottoMachine = LottoMachine(winningNumbers, bonusNumber)
+        val machine = LottoMachine(winningNumbers, bonusNumber)
         val resultTable = machine.createResultTable(allTickets)
 
-        val winningMoney = WinStatCalculator.calculateWinningMoney(resultTable)
-        val returnRate = WinStatCalculator.calculateReturnRate(winningMoney, amountOfMoney)
+        val winStatCalculator = WinStatCalculator()
+        val winningMoney = winStatCalculator.calculateWinningMoney(resultTable)
+        val returnRate = winStatCalculator.calculateReturnRate(winningMoney, amountOfMoney)
 
         ResultView.displayFinalResults(resultTable, returnRate)
     }
--- a/src/main/kotlin/lotto/Lotto.kt
+++ b/src/main/kotlin/lotto/Lotto.kt
@@ -3,23 +3,29 @@ package lotto
 object Lotto {
     fun run() {
         val amountOfMoney = retryable { InputView.getPurchaseAmount() }
-        val printer = LottoPrinter(amountOfMoney)
+        val totalTickets = amountOfMoney / LottoTicket.COST_OF_TICKET
 
-        ResultView.displayNumberOfTickets(printer.amountOfTicket)
-        ResultView.displayTickets(printer.bundleOfLottoTicket)
+        val manualCount = retryable { InputView.getManualTicketCount(totalTickets) }
+        val manualTickets = retryable { InputView.getManualTickets(manualCount) }
 
-        val lastWeekWinningNumbers = retryable { InputView.getLastWeekWinningNumbers() }
+        val autoCount = totalTickets - manualCount
+        val autoTickets = LottoPrinter.generateLottoTickets(autoCount)
+
+        val allTickets = manualTickets + autoTickets
+
+        ResultView.displayTicketPurchaseSummary(manualCount, autoCount)
+        ResultView.displayTickets(allTickets)
+
+        val winningNumbers = retryable { InputView.getLastWeekWinningNumbers() }
         val bonusNumber = retryable { InputView.getBonusNumber() }
-        println()
 
-        val machine = LottoMachine(amountOfMoney, lastWeekWinningNumbers, bonusNumber)
-        machine.bundleOfLottoTicket = printer.bundleOfLottoTicket
-        machine.writeResultTable()
+        val machine = LottoMachine(winningNumbers, bonusNumber)
+        val resultTable = machine.createResultTable(allTickets)
 
-        val winningMoney = WinStatCalculator.calculateWinningMoney(machine.resultTable)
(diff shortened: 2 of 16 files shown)


src/main/kotlin/lotto/Lotto.kt
```diff
@@ -3,23 +3,29 @@ package lotto
...
+
+        val winningNumbers = retryable { InputView.getLastWeekWinningNumbers() }
         val bonusNumber = retryable { InputView.getBonusNumber() }
-        println()
 
-        val machine = LottoMachine(amountOfMoney, lastWeekWinningNumbers, bonusNumber)
-        machine.bundleOfLottoTicket = printer.bundleOfLottoTicket
-        machine.writeResultTable()
+        val machine = LottoMachine(winningNumbers, bonusNumber)
+        val resultTable = machine.createResultTable(allTickets)
```
Wordbe, line 24: For variables where the type isn’t immediately clear, it’s a good idea to explicitly specify the type - it helps improve readability and understanding! 😊

src/main/kotlin/lotto/Lotto.kt
```diff
@@ -3,23 +3,29 @@ package lotto
...
 
-        val winningMoney = WinStatCalculator.calculateWinningMoney(machine.resultTable)
-        val returnRate = WinStatCalculator.calculateReturnRate(winningMoney, machine.amountOfMoney)
+        val winningMoney = WinStatCalculator.calculateWinningMoney(resultTable)
+        val returnRate = WinStatCalculator.calculateReturnRate(winningMoney, amountOfMoney)
 
-        ResultView.displayWinningStatistics(machine.resultTable)
+        ResultView.displayWinningStatistics(resultTable)
```
Wordbe: Instead of calling two methods separately, it would be cleaner to call returnRate once and handle everything together in that call! 😊

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,20 +1,15 @@
...
-    val resultTable: MutableMap<Rank, Int> = mutableMapOf()
-
-    fun writeResultTable() {
-        bundleOfLottoTicket.forEach { filterTicket(it) }
-    }
-
-    fun filterTicket(ticket: LottoTicket) {
-        val rank = Rank.ofTicket(ticket, lastWeekWinningNumbers, bonusNumber)
-        resultTable[rank] = resultTable.getOrDefault(rank, 0) + 1
+    fun createResultTable(tickets: List<LottoTicket>): Map<Rank, Int> {
```
Wordbe: How about replacing the Map with a dedicated object that better represents its purpose?
While the key type Rank is easy to understand, the value’s meaning isn’t immediately clear.

Using a well-named class instead can improve readability, maintainability, and make the domain intent more explicit! 😊

src/main/kotlin/lotto/LottoPrinter.kt
```diff
@@ -1,11 +1,8 @@
 package lotto
 
-class LottoPrinter(val amountOfMoney: Int) {
-    val amountOfTicket: Int = amountOfMoney / LottoTicket.COST_OF_TICKET
-    val bundleOfLottoTicket: List<LottoTicket> = generateLottoTickets(amountOfTicket)
-
-    fun generateLottoTickets(amountOfTicket: Int): List<LottoTicket> {
-        return List(amountOfTicket) {
+object LottoPrinter {
```
Wordbe: For domain-related objects, instead of using object with static-like methods, it's better to design them so that dependencies are injected from the outside in.
This approach improves flexibility and makes your code much easier to test! 😊

Take a moment to review your other object declarations as well -
consider whether they could be refactored into classes with injected dependencies.

src/main/kotlin/lotto/LottoMachine.kt
```diff
@@ -1,20 +1,15 @@
 package lotto
 
 class LottoMachine(
-    val amountOfMoney: Int,
-    val lastWeekWinningNumbers: List<Int>,
-    val bonusNumber: Int,
+    private val lastWeekWinningNumbers: List<Int>,
+    private val bonusNumber: Int,
```
Wordbe: Try wrapping the primitive value in a class —
it’s a great opportunity to explore the benefits, such as improved type safety, clearer intent, and the ability to add related behavior! 😊

src/main/kotlin/lotto/ManualTicketParser.kt
```diff
@@ -0,0 +1,17 @@
+package lotto
+
+object ManualTicketParser {
```
Wordbe: How about organizing your packages into view, domain, and controller layers?
Then consider where ManualTicketParser fits best.

Thinking this way helps clarify roles and improves maintainability! 😊

Wordbe (changes requested):
Hello! Sari
It looks like you’ve done a great job implementing the requirements for step 2 - thank you!

I’ve left some comments with things that might be worth thinking about.
Please take a look and apply any changes you find helpful.
And of course, if you have a different opinion, feel free to share it!

Thanks again! 😊

src/main/kotlin/lotto/Lotto.kt
saragrosser: I have refactor this to add explicit type declarations for readability, I will keep this in mind. Thank you!!

src/main/kotlin/lotto/Lotto.kt
saragrosser: Thank you! I have refactor and combine result display methods into a single call this makes much more sense :)

src/main/kotlin/lotto/LottoMachine.kt
saragrosser: Hi Wordbe, thank you! I created "BonusNumber" and "WinningNumbers" as value classes to replace raw Int and List<Int>. I am slowly understanding better the domain modeling in Kotlin. :)

src/main/kotlin/lotto/LottoMachine.kt
saragrosser: Thanks Wordbe, I’ve replaced the raw Map<Rank, Int> with a new WinningStatistics class. This class now encapsulates the rank-counting logic 

To apply the change, I updated the LottoMachine "createResultTable" function to return a WinningStatistics object instead of a map. I also updated the related test cases and adapted WinStatCalculator to work with the new class by using countOf(rank) internally.

src/main/kotlin/lotto/ManualTicketParser.kt
saragrosser: Thank you so much for your feedback I will defenitely consider this for my following missions, because I am running out of time for the next blackjack project deadlines I will for now leave it as it is and come back to it later when I have less time constrains if that is okey, thank you!

src/main/kotlin/lotto/LottoPrinter.kt
saragrosser: Thank you for your feedback, I have to study more to avoid these type of mistakes, im still unsure when to use object or class, but for now have I refactored LottoPrinter from an object to a class and created the instance insite Lotto.kt :)

src/main/kotlin/lotto/BonusNumber.kt
```diff
@@ -0,0 +1,3 @@
+package lotto
+
+data class BonusNumber(val value: Int)
```
Wordbe, line 3: Why did you use a data class? 😊

Wordbe: Try creating a LottoNumber class in addition to BonusNumber, and ensure that for numbers 1 to 45, the same instance is always returned per number. 😊

src/main/kotlin/lotto/WinningNumbers.kt
```diff
@@ -0,0 +1,3 @@
+package lotto
+
+data class WinningNumbers(val numbers: List<Int>)
```
Wordbe, line 3: Why did you use `data class` here? 🙂

Wordbe (approved):
Hello Sari!
Thank you for consistently refining and completing the Lotto mission.

Since you’ve reflected most of the feedback I gave, I’ll go ahead and approve and merge the PR!

Although you must be busy with the Blackjack mission as well, I’d like to share a few suggestions for further improvement, in addition to the comments I’ve already left:

Try structuring your code with a controller, model, and view separation. Consider collecting LottoTicket instances into a first-class collection. Reflect on the comments and try applying them further. 

Great job, and thank you for putting your best into the Lotto mission!
I hope what you’ve learned and explored here will continue to help you write better code moving forward. 😊

PR #31, opened 2025-07-08
[Lotto Step 2] Submit pebencze's mission

pebencze:
Hello!

I have implemented additional feature to my lotto game: the user now can input a certain number of tickets manually. The rest of the tickets is generated automatically. Even though it is not many new features, I struggled a bit with creating classes and avoiding unnecessary dependencies and feel like my LottoHandler class is spaghetti. However, I had to move on to the next project and decided to submit, even though I have a few ideas on how to improve. I would appreciate some feedback on the followings:
Q1: How shall I store tickets, manualTickets and automaicTickets?
Q2: Where shall I use first class collections and wrappers for primitives? E.g. instead of List<Ticket> - Tickets, instead of Int - Money?
Q3: How is my purchase class?
Q4: Should I use inheritance and composition anywhere?

Thank you so much for reviewing my code!


src/main/kotlin/lotto/Lotto.kt
```diff
@@ -1,25 +1,29 @@
...
-        require(numbers.all { it in MIN..MAX }) {
-            "[ERROR] Range has to be $MIN..$MAX."
-        }
         require(numbers.distinct().size == 6) {
             "[ERROR] Duplicates found."
         }
     }
+    val sortedNumbers = numbers.sortedBy { it.number }
+
+    override fun toString(): String {
```
Wordbe: It looks like the Lotto class contains something intended for the View layer.
Where do you think would be a better place for that?

Consider whether it truly belongs in the domain, or if it would be more appropriate in the View or Controller layer — keeping a clear separation of concerns can lead to a cleaner architecture! 😊

src/main/kotlin/lotto/LottoNumber.kt
```diff
@@ -0,0 +1,14 @@
+package lotto
+
+class LottoNumber private constructor(val number: Int) {
+    companion object {
+        const val MIN = 1
+        const val MAX = 45
+        private val NUMBERS: Map<Int, LottoNumber>
```
Wordbe, line 7: You’ve implemented a strategy to limit unbounded instance creation - that’s a great approach! 😊

src/main/kotlin/lotto/LottoHandler.kt
```diff
@@ -1,31 +1,77 @@
 package lotto
 
 object LottoHandler {
+    private val machine = LottoMachine()
+    private val tickets: List<Lotto>
```
Wordbe: Please remove any unnecessary state from within LottoHandler.
Keeping it stateless where possible will make the code simpler, more predictable, and easier to test! 😊

The tickets-related variable could be retrieved on demand instead of being stored.

src/main/kotlin/lotto/Purchase.kt
```diff
@@ -0,0 +1,26 @@
+package lotto
+
+data class Purchase(val amount: Int) {
```
Wordbe: Why did you use a data class? Is it really necessary? 🤔

src/main/kotlin/lotto/Purchase.kt
```diff
@@ -0,0 +1,26 @@
...
+    init {
+        require(amount in MIN..MAX) {
+            "[ERROR] Max purchase amount allowed is $MIN-$MAX."
+        }
+    }
+
+    val change = amount % TICKET_PRICE
```
Wordbe: There are several state variables here. Instead, try creating methods that return values when needed.

Also, implement the TODOs if necessary. 😊

For reference, a value class can be used when there's a single property - it wraps a primitive type to clearly express the purpose of the object.

src/main/kotlin/lotto/Statistics.kt
```diff
@@ -1,35 +1,27 @@
...
+    purchaseAmount: Int,
     private val winningNumbers: WinningNumbers,
-    private var _results: MutableMap<Rank, Int> = emptyMap<Rank, Int>().toMutableMap(),
 ) {
-    init {
-        calculateResults(tickets)
-    }
-    val results: Map<Rank, Int>
-        get() = _results
+    val results: Map<Rank, Int> = calculateResults(tickets)
```
Wordbe: Instead of creating internal variables, return the result directly through a method. 😊

Wordbe: It’s hard to understand what the value in the Map represents. How about creating an object that holds both Rank and Int to represent this more clearly? 😊

src/test/kotlin/lotto/LottoNumberTest.kt
```diff
@@ -0,0 +1,27 @@
+package lotto
+
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+import org.junit.jupiter.api.assertThrows
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ValueSource
+
+class LottoNumberTest {
```
Wordbe, line 10: Great test 👍

Wordbe (changes requested):
Hello Petra!
Thank you for your thoughtful code and all the effort you put into the submission.

I've answered your question below, so please check it out.
Also, take a look at the review comments and apply any suggestions you find helpful.

Thank you! 😊

---

Q1: How shall I store tickets, manualTickets and automaicTickets?

It would be cleaner to compute the needed value as a local variable within the block. 😊

Q2: Where shall I use first class collections and wrappers for primitives? E.g. instead of List - Tickets, instead of Int - Money?

That’s right. You can turn List<Lotto> into a first class collection, and you can also wrap Int in a Money object.
Trying this out will help you get more comfortable working with objects effectively! 😊

Q3: How is my purchase class?

I’ve left a review. please check the comments!

Q4: Should I use inheritance and composition anywhere?

There are cases where inheritance is appropriate, and others where composition is better.
I recommend exploring the pros and cons of each and trying them out in practice. 😊

---

Thank you again.

src/main/kotlin/lotto/Lotto.kt
pebencze: True! Thank you for the feedback! I tried moving it to the View.

src/main/kotlin/lotto/LottoHandler.kt
pebencze: Will try to do that. Doing the next project I have discovered that it is better to avoid state in certain classes. Or would you say it is good practice in general?

src/main/kotlin/lotto/Purchase.kt
pebencze: Hm. I was experimenting, but I think I will need to read more about this topic. I wanted to try things and use value and data classes, but I could not find a right fit.

src/main/kotlin/lotto/Purchase.kt
pebencze: How do you find the new changes? I feel like my Handler became a bit bloated, but the other classes are simpler.

src/main/kotlin/lotto/Statistics.kt
pebencze: Makes sense! Thank you.

pebencze: Done, thank you!

src/main/kotlin/lotto/LottoHandler.kt
pebencze: What do you think of my new approach in the Handler class? :) It became a bit long, didn't it?

pebencze:
Hello [Wordbe](https://github.com/Wordbe)!

Thank you for reviewing my code fast & thoroughly!

I have tried to implement most of your feedback and I think I will be able to use them in the future. Many things make more sense to me now!

However, there are some points I still struggle with and would appreciate some comments on, but I will also read more documentation on them and implement them in other projects.
- P1: implementing the MVC pattern creates a bloated Controller class. I hope to improve that in future projects.
- P2: first-class collections make a lot of sense, but they also result in an extra "." when accessing the value and calling methods on them.
- P3: Still a bit confused when choosing between: methods (stateless class) vs. properties with getters and setters (state). Is there a "golden rule" for that?

I appreciate your review and hope that my refactoring has worked.

All the best,
Petra

README.md
```diff
@@ -1,5 +1,12 @@
 # kotlin-lotto
 ## Features
+
+### General
+- [ ] Wrap primitives
```
Wordbe: Check this please!🙂

src/main/kotlin/lotto/LottoHandler.kt
```diff
@@ -3,41 +3,89 @@ package lotto
...
+                println(err.message)
+            }
+        }
+        throw IllegalArgumentException(MAX_ATTEMPT_MESSAGE)
+    }
+
+    fun buyManualTickets(ticketCount: Int): List<Lotto> {
+        repeat(MAX_ATTEMPT) {
+            try {
+                val tickets = mutableListOf<Lotto>()
```
Wordbe: Don’t declare variables in advance—create them only when needed!
Also, you can create a list directly like listOf("a").
Try to use immutable lists whenever possible. 😊

src/main/kotlin/lotto/LottoHandler.kt
Wordbe: Yes, looks good! Although it's long, it's much easier to read now thanks to the well-divided methods. Please mark methods that are used only internally as private. 😊

src/main/kotlin/lotto/Purchase.kt
Wordbe: Sure! Please look deeper into the difference between data class and value class.

In this case, value class seems more appropriate, as it’s well-suited for wrapping a single primitive property to represent a domain model. It also comes with the bonus of no runtime wrapping overhead, which means better performance.

Give it a try! 🙂

src/main/kotlin/lotto/Purchase.kt
Wordbe: Your hanlder sounds good!
But please remove the state variables I mentioned above (change, ticketCount).
You don't need to store them—just return them as method results instead!

src/main/kotlin/lotto/Statistics.kt
Wordbe: Got it!
I wasn’t suggesting to put a Map inside Result, but rather to create rank and count properties—so that the meaning of the key and value is clear. What do you think about that? 😊

Wordbe (changes requested):
Thank you for thoughtfully reflecting on the feedback, Petra.
Here's a review in response—please take a look!

Also, I’d like to share my thoughts on the areas you’ve been considering. 😊

---

P1: implementing the MVC pattern creates a bloated Controller class. I hope to improve that in future projects.

If you encapsulate core logic within the domain model and create abstracted methods, the controller can simply call those methods, reducing the amount of code in the controller. It may be challenging, but I hope you'll consider giving it a try!!

P2: first-class collections make a lot of sense, but they also result in an extra "." when accessing the value and calling methods on them.

I also think that’s one of the drawbacks of an object wrapper collection. Instead of exposing getters only when absolutely necessary, if you create the needed methods directly inside the collection class, you can reduce the use of “.” calls!
I hope you try creating a first-class collection in this Lotto mission. 

P3: Still a bit confused when choosing between: methods (stateless class) vs. properties with getters and setters (state). Is there a "golden rule" for that?

Think about someone else reading this code. Code with many internal states (variables) can make it hard to quickly understand where and how those states change or are used. That’s why it's better to avoid unnecessary state in domain objects, keeping only the essential state. This also helps reduce potential side effects caused by mutable state. What do you think about this, Petra? 🙂

---

Thank you again for giving your best effort on the mission!

src/main/kotlin/lotto/Statistics.kt
pebencze: I see. That would make sense, too. The problem is that if I make Count a value class, I cannot modify and therefore increment it. Would it then become a data class?

pebencze:
Hello Wordbe!

Thanks for the feedback!

I have implemented the requested changes and the code looks better to me. Also, I have more clarity about value classes and the MVC pattern now. I would leave Result as it is for now and consider different wrappings in future projects. I have also left a comment on that.

Would you please review the changes?

Best,

Petra

src/main/kotlin/lotto/Statistics.kt
Wordbe: What I meant was a class that has rank and count as its properties.
For example, you could create a Result object that contains rank and count.

Making Count a value class is a separate matter.
If you want to create an object with an incremented value, creating Count(2) instead of Count(1) should be perfectly fine. 😊

Wordbe (approved):
Hello Petra,

Thank you for thoughtfully considering the feedback and diligently applying it.

We'll wrap up the Lotto mission here — I'm approving and merging it.

Please take a moment to reflect on both the previous explanations and the new comments I've left. It would be great if you could incorporate them when you have time.

I hope this Lotto mission provides a meaningful stepping stone for your future coding journey. Thank you again! 🙂🙂

